PR #29, opened 2021-04-22
Offer additions and transfer string change

nategeier:


src/wallet.ts
```diff
@@ -335,7 +335,7 @@ export class Wallet {
         account_id: MARKET_ACCOUNT,
         msg: JSON.stringify({
           price: price,
-          autotransfer: autotransfer || true,
```
nategeier, line 338: if sent in false, it would always be true, (no false = true)

src/wallet.ts
```diff
@@ -461,18 +461,82 @@ export class Wallet {
...
+      price
+    )
+  }
+
+  /**
+   * Make an offer to a token.
+   * @param tokenId
+   * @param price
+   */
+  public async acceptAndTransfer(
```
nategeier, line 513: this function is the final pull from the lister and transfers the NFT, works like a charm, thanks @thor314

src/wallet.ts
microchipgnu: 👍

microchipgnu (approved):

PR #39, opened 2021-05-04
Error handling refactor

microchipgnu:


__tests__/unit/file-utils.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  //     const file = new File([''], name, { type: 'text/plain' })
+
+  //     try {
+  //       await correctFileType(file)
+  //     } catch (error) {
+  //       throw new Error(error.message)
+  //     }
+  //   }).toThrow(ERROR_MESSAGES.fileNoExtension)
+  // })
```
jpcloureiro, line 51: tip: you can do `test.skip( ... ) instead of commenting out.

This is helpful because jest outputs how many tests were skipped and we can track them down if we want to handle them later in time.

package.json
```diff
@@ -77,5 +78,9 @@
     "typedoc": "^0.20.33",
     "typescript": "^4.2.4",
     "uglifyify": "^5.0.2"
+  },
+  "jest": {
+    "verbose": true,
+    "testEnvironment": "jsdom"
```
jpcloureiro, line 84: You were right, `testEnvironment` config here doesn't have any effect.

jpcloureiro (approved):
looks good to me! :shipit:

__tests__/unit/file-utils.test.ts
microchipgnu: Nice tip! 🎉

package.json
microchipgnu: Yeah, I'm keeping it and until we find an alternative.

Adding ` /* @jest-environment jsdom */` in the beginning of our test file seems to do the trick.

PR #40, opened 2021-05-05
Dynamic list cost

microchipgnu:
Removes `LIST_COST` constant and adds a helper that calculates the storage cost for a market listing.


jpcloureiro (approved):
looks good!

ship it :shipit:

PR #50, opened 2021-05-10
Externalize constants

microchipgnu:
This PR changes fundamentally how we deal with constants. We now handle constants externally in a versioned way.

For wallet initialization, you need to `await new Wallet.init({...walletConfig})` now. This will allow valid API Key owners to fetch constants (current factory contract name and a few other constants from our backend). Once there's a response, it will pass these constants down to the API, Storage and Minter instances. 

A new class was created for Storage. This means Storage can also be initialized and used on its own. 

Tests now run on npm publishing.

### DIFF
--- a/src/minter.ts
+++ b/src/minter.ts
@@ -121,6 +121,8 @@ export class Minter {
     error: null | string
   }> {
     try {
+      if (!this.storage) throw new Error('Storage not initialized')
+
       // corrects MIME type.
       const tFile = await correctFileType(file)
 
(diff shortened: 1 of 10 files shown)


nategeier (changes requested):

src/minter.ts
```diff
@@ -112,13 +124,23 @@ export class Minter {
...
 
-      if (tFile.size > FILE_UPLOAD_SIZE_LIMIT)
+      if (
+        tFile.size >
+        (this.constants.FILE_UPLOAD_SIZE_LIMIT || FILE_UPLOAD_SIZE_LIMIT)
+      )
         throw new Error(ERROR_MESSAGES.fileSizeExceeded)
 
-      const result = await uploadToArweave(file, this.apiKey)
+      if (!this.storage) throw new Error('Storage not initialized')
```
jpcloureiro, line 133: Should check storage instance initialization before filesize check

jpcloureiro (changes requested):

src/minter.ts
microchipgnu: 👍

PR #62, opened 2021-05-14
fix undefined on fetchThingMetadata

### LINKED ISSUE #61: TypeError: Cannot read property '0' of undefined
https://github.com/Mintbase/mintbase-js/blob/4a5906fde6d692b35fa120f6d7e6586e8267742d/src/api.ts#L143

Should be `result.thing.tokens[0]`

jpcloureiro:
fixes #61

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -7,7 +7,7 @@ import {
   Network,
   Chain,
   Constants,
-  ThingMetadata,
+  MintMetadata,
 } from './types'
 import { API_BASE_NEAR_MAINNET, API_BASE_NEAR_TESTNET } from './constants'
 import { formatResponse, ResponseData } from './utils/responseBuilder'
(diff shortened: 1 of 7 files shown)


src/api.ts
```diff
@@ -140,7 +140,7 @@ export class API {
     if (result.thing.length === 0)
       return formatResponse({ error: `${thingId} is not a valid thing.` })
 
-    const thing = result.token[0]
+    const thing = result.thing.token[0]
```
microchipgnu, line 143: I think it is `result.thing.tokens[0]`

microchipgnu (changes requested):

src/types.ts
```diff
@@ -182,6 +182,30 @@ interface MintMetadata {
...
+  split_revenue: null
+  tags: string[]
+  media: string
+  extra: []
+  title: string
+  store: string
+  external_url: string
+  type: string
+}
+
```
microchipgnu, line 208: Why not reuse `MetadataField`?

microchipgnu (changes requested):

src/types.ts
jpcloureiro: 👀

microchipgnu (approved):

PR #79, opened 2021-06-01
Fix royalties args on Mint More

microchipgnu:
Minting more tokens of when royalties were set to null resulted in badly formatted args. Fixed it.

### DIFF
--- a/src/wallet.ts
+++ b/src/wallet.ts
@@ -915,16 +915,15 @@ export class Wallet {
         STORE_CONTRACT_CALL_METHODS,
     })
 
-    let _royalties: any = {}
-
-    token.royaltys.forEach((royalty) => {
-      _royalties = {
-        ..._royalties,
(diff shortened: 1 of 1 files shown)


src/wallet.ts
```diff
@@ -915,16 +915,15 @@ export class Wallet {
...
-    if (Object.keys(_royalties).length === 0) _royalties = null
+    const _royalties = token.royaltys.reduce(
+      (accumulator, { account, percent }) => {
+        return {
+          ...accumulator,
+          [account]: percent,
+        }
+      },
+      {}
+    )
```
jpcloureiro, line 926: 👌

jpcloureiro (approved):
Great improvement on the royalties obj build

looks good. ship it.

src/wallet.ts
microchipgnu: 🙏

PR #119, opened 2022-04-19
Enable optional method arguments for all contract calls

microchipgnu:
**Highlights**
1. Update NEAR API JS to the latest version (0.44.2)
2. Creates type `OptionalMethodArgs` that is used optionally on all methods
3. Enabled setting `timeout` when making offers to any arbitrary integer greater than 24

```ts
type OptionalMethodArgs = {
  gas?: string
  amount?: string
  meta?: string
  callbackUrl?: string
}
```


sainthiago (approved):

src/wallet.ts
```diff
@@ -339,11 +349,13 @@ export class Wallet {
...
-      { token_ids: tokenIds },
-      MAX_GAS,
-      ONE_YOCTO
-    )
+    await contract.nft_batch_transfer({
+      meta: 'nft_batch_transfer',
+      callbackUrl: options?.callbackUrl,
+      args: { token_ids: tokenIds },
+      gas: gas,
+      amount: ONE_YOCTO,
```
tifrel, line 357: So we willfully ignore anything that a user of mintbase-js would supply as amount? Contract panics on anything else than 1 yocto, but I consider it better not to silently override user input.

src/wallet.ts
```diff
@@ -339,11 +349,13 @@ export class Wallet {
     })
 
     // @ts-ignore: method does not exist on Contract type
-    await contract.nft_batch_transfer(
-      { token_ids: tokenIds },
-      MAX_GAS,
-      ONE_YOCTO
-    )
+    await contract.nft_batch_transfer({
+      meta: 'nft_batch_transfer',
```
tifrel, line 353: See below, we just silently override user input?

microchipgnu: Good point, will remove these.

src/wallet.ts
microchipgnu: Yes only for the ones we know the contract panics.

sainthiago (approved):

src/wallet.ts
```diff
@@ -907,7 +975,14 @@ export class Wallet {
       : new BN(options?.attachedDeposit)
 
     // @ts-ignore: method does not exist on Contract type
-    await contract.create_store(storeData, gas, attachedDeposit)
+    await contract.create_store({
```
cif, line 978: It would be a lot nicer if the method did exist on contract type :) Like @tifrel idea of generating these interfaces from Rust code

cif (approved):
Looks good to me, but let's definitely do a full manual regression test before pushing live

src/wallet.ts
microchipgnu: I think part of this limitation comes from using the `Contract` type from `near-api-js`, but 100% on figuring out a solution. Generating from the source would be awesome.

PR #121, opened 2022-05-05
🐛 Add reference hash and others to metadata arguments

microchipgnu:


sainthiago (approved):

tifrel (changes requested):

src/wallet.ts
```diff
@@ -1087,9 +1089,13 @@ export class Wallet {
...
       owner_id: accountId,
       metadata: {
+        title: metadata?.title,
+        description: metadata?.description,
+        media: metadata?.media,
+        media_hash: metadata?.media_hash,
         reference: metadataId,
-        // TODO: check if category is lowercase
-        extra: !category ? null : category,
+        reference_hash: metadataId,
```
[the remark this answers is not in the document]
microchipgnu: Care to clarify?

cif (approved):

src/utils/crypto.ts
```diff
@@ -0,0 +1,10 @@
+const hash = async (buffer: ArrayBuffer) => {
+  const hashBuffer = await crypto.subtle.digest('SHA-256', buffer)
+  const hashArray = Array.from(new Uint8Array(hashBuffer))
+  const base64 = btoa(
+    hashArray.map((b) => b.toString(16).padStart(2, '0')).join('')
```
tifrel, line 5: Interesting, why the base16 in the middle?

src/wallet.ts
```diff
@@ -1087,9 +1090,13 @@ export class Wallet {
     const obj = {
       owner_id: accountId,
       metadata: {
+        title: metadata?.title,
+        description: metadata?.description,
+        media: metadata?.media,
+        media_hash: metadata?.media_hash,
```
microchipgnu, line 1096: Shall we keep these here?

@tifrel @cif

tifrel (approved):

src/wallet.ts
tifrel: Why not. But we need to be sure that if media is set, media_hash is set as well

microchipgnu: Yes, when uploading through MintbaseJS `media_hash` is always generated.

cif:
Closing this, we'll revisit as we make refactoring changes to the minter

PR #124, opened 2022-06-06
:sparkles: Add Batch Change Minters

sainthiago:


src/wallet.ts
```diff
@@ -1266,6 +1266,44 @@ export class Wallet {
...
+      callbackUrl: options?.callbackUrl,
+      args: {
+        grant: grantMinters.length > 0 ? grantMinters : undefined,
+        revoke: revokeMinters.length > 0 ? revokeMinters : undefined,
+      },
+      gas: gas,
+      amount: ONE_YOCTO,
+    })
+
+    // TODO: define a response for this
```
cif, line 1303: Does the client need this?  I do wonder how we'll want to adapt to clients that don't require redirects (e.g. Sender) in the future

cif (approved):

microchipgnu (approved):

PR #126, opened 2022-06-07
Adding media to on-chain metadata

tifrel:


src/wallet.ts
```diff
@@ -1093,8 +1096,9 @@ export class Wallet {
       owner_id: accountId,
       metadata: {
         reference: metadataId,
-        // TODO: check if category is lowercase
-        extra: !category ? null : category,
+        media: mediaId,
```
cif, line 1099: Will mediaId be the full URL here?  For example https://my3n42vh6gurlncgfouttpy6ai3bqn6b7idmvvjukfkzotmvoe.arweave.net/ZjbeaqfxqRW0RiupOb8eAjYYN8H6BsrVNFFVl02_VcU returns full url in "media" field.  Maybe we should strip out the base uri to keep the contract.base_uri + token.meta.media intact?

tifrel:
As decided per sprint dev sync, not gonna continue this for now.

PR #141, opened 2022-08-24
♻️ Migrate thing query

rubenmarcus:
@microchipgnu  I see an issue on the royalties prop, the object that comes isn't the same of the old query is it?

I left the type the same as it was, but it will not have issues to the users using the wallet?

`  royalties: { account: string; percent: string }[]`

### DIFF
--- a/src/wallet.ts
+++ b/src/wallet.ts
@@ -1152,7 +1152,7 @@ export class Wallet {
         royalties: Record<string, number>[]
       }[]
     }>(
-      `query GET_THING_BY_ID($id: String!) {
+      `query v2_getByThingId($id: String!) {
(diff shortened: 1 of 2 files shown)


src/wallet.ts
```diff
@@ -1148,23 +1148,17 @@ export class Wallet {
         metaId: string
         storeId: string
         memo: string
-        tokens: {
-          royaltyPercent: string
-          royaltys: { account: string; percent: string }[]
-        }[]
+        royalties_percent: string
+        royalties: { account: string; percent: string }[]
```
microchipgnu, line 1152: I think the `royalties` type is now `Record<string, number> `

rubenmarcus: Done

tifrel (approved):

microchipgnu (approved):

PR #154, opened 2022-10-03
✨ Support wallet based transactions

cif:
Add core call wrapper around wallet selector
Expose getWallet method
React experiment (simple transfer call) 
Tests 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203089260354071


packages/sdk/src/calls.ts
```diff
@@ -0,0 +1,63 @@
...
+export type NearCallSigningOptions = {
+  wallet?: Wallet;
+  account?: Account;
+};
+
+export class NoSigningMethodPassed extends Error {
+  message: string;
+}
+
+export const callContractMethodWithWallet = async (
```
cif, line 24: Should probably not export this

nategeier (commented):
Excelent

microchipgnu:
Merging this!

PR #157, opened 2022-10-27
🗃️ Add data module

cif:
This adds a @mintbase-js/data package to the monrepo. 

Generates graphql types via graphql-codegen
Provides wrapper example for getting listing counts for a given meta id
For discussion (see comment).  

Used @rubenmarcus wrapper on omni site and personally enjoy the `const { error, data } = await getThings()`  so maybe we want to continue that pattern in the helpers vs log and re throw.


packages/data/src/graphql/tokenListingCounts/tokenListingCountsByMetaId.ts
```diff
@@ -0,0 +1,55 @@
...
+    query: tokenListingCountsByMetaIdQuery,
+    variables: {
+      metadataId,
+    },
+  });
+
+  // log and rethrow for now...
+  // question: do we want to use the same { error, data } pattern in these
+  // methods as well?
+  if (error) {
```
cif, line 33: See the description for conversation around this.  Almost wondering if we offer two methods, one which will throw and one that will offer the error, data result shape 🤔

packages/data/src/graphql/tokenListingCounts/tokenListingCountsbyMetaId.integration.test.ts
```diff
@@ -0,0 +1,10 @@
+import { tokenListingCountsByMetaId } from './tokenListingCountsByMetaId';
+
+const METADATA_ID_FOR_TEST = 'run.mintbase1.near:7517c09a546d39cdbbd12e07a8c1de8f';
+
+describe('tokenListingCountsByMetaId', () => {
+  it('should return null for now', async () => {
```
cif, line 6: Need to update this

microchipgnu (approved):

PR #158, opened 2022-11-02
feat: storage module

microchipgnu:


cif:
Does this use the new storage SDK that we got surprised with recently?

PR #159, opened 2022-11-02
feat: add calls implementation for token and market contract

microchipgnu:


cif:
Odd the build fails on this one, wondering if the tsc package order matters, it does appear to be trying to compile react first

packages/sdk/src/mb-impl/constants.ts
```diff
@@ -0,0 +1,41 @@
...
+  'TOKEN_ACCOUNT_REVOKE_ALL' = 'nft_revoke_all',
+}
+
+export enum MARKET_METHOD_NAMES {
+  'LIST' = 'nft_approve',
+  'DEPOSIT_STORAGE' = 'deposit_storage',
+  'BUY' = 'buy',
+}
+
+export enum GAS_CONSTANTS {
```
cif, line 21: Why not just use regular constant string values for these?  Enumerated objects are good for referencing values by index (number) in a list form, but purely for organization/collection cases where only a few values are needed or just using strings in general seems overkill

cif: *Applies to other enums in here as well I think, they aren't really lists of a single field

packages/react/src/hooks/useMinter.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  const [loading, setLoading] = useState(false);
+
+  const handleMint = async (): Promise<void> => {
+    setLoading(true);
+
+    // TODO: use @mintbase/storage to get a reference
+    const reference = '';
+    
+    await callContractMethod(
```
cif, line 27: do we want to borrow @tifrel 's `execute` name?  I don't love `callContractMethod` but it is rather explicit

packages/sdk/src/mb-impl/token.ts
```diff
@@ -0,0 +1,214 @@
...
+  RevokeAccountArgs,
+} from './token.types';
+
+// TODO: how to deal with this differently?
+const getFactoryContract = (network: string): AccountId => {
+  if (network === Network.MAINNET) return MB_MAINNET_TOKEN_FACTORY;
+  else return MB_TESTNET_TOKEN_FACTORY;
+};
+
+// TODO: figure out a way to generate gas and deposit for each
```
cif, line 30: 👍

packages/sdk/src/mb-impl/token.ts
```diff
@@ -0,0 +1,214 @@
...
+  DeployTokenContractArgs,
+  AccountId,
+  TransferTokenContractOwnership,
+  MintArgs,
+  AddRemoveMinterArgs,
+  BatchChangeMinters,
+  RevokeAccountArgs,
+} from './token.types';
+
+// TODO: how to deal with this differently?
```
cif, line 24: This type of method only become necessary in cases where the network can be switched at runtime, so is there a way to avoid this? e.g. require a refresh for browser network switches and do something like:

```
const NEAR_NETWORK = process.env.NEAR_NETWORK || global.NEAR_NETWORK || localStorage?.get('NEAR_NETWORK')
 ```
 
If we can prevent this, network will be set at environment/browser level vs switched at runtime I'm okay with setting constants based on that e.g. 
```
const TOKEN_CONTRACT_FACTORY = NEAR_NETWORK === 'mainnet' 
    ?  TOKEN_CONTRACT_FACTORY_MAINNET
    : TOKEN_CONRACT_FATORY_TESTNET
```

We can then just omit this super basic logic from test coverage (constants are omitted)

File: packages/sdk/src/index.ts [hunk 5]
```diff
@@ -1,2 +1,4 @@
 export * from './calls';
 export * from './constants';
+export * from './mb-impl/token';
```
cif, line 3: Why have sub dir for our implementation? Guess my thought there is its already ours, maybe a goos set of functions to version though? e.g `sdk/src/v1`?

packages/react/src/hooks/useMinter.ts
```diff
@@ -0,0 +1,54 @@
...
+      {
+        ...mint({
+          nftContractId,
+          metadata: {
+            reference: reference,
+          },
+          options,
+        }),
+        gas: MAX_GAS,
+        deposit: '0',
```
tifrel, line 37: Needs to be one yocto

packages/react/src/hooks/useMinter.ts
tifrel: I would actually be a fan of `view` and `exec` (or `call`, but I find that too generic) to signal a difference between mutations and pure information retrieval

packages/sdk/src/mb-impl/market.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  return {
+    contractAddress: nftContractId,
+    args: {
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      token_id: tokenId,
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      account_id: approvedAccountId,
+      msg: JSON.stringify({
+        price: price.amount,
```
tifrel, line 45: Aren't we missing the currency here?

packages/sdk/src/mb-impl/market.ts
```diff
@@ -0,0 +1,67 @@
...
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      nft_contract_id: nftContractId,
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      token_id: tokenId,
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      referrer_id: referrerId,
+    },
+    methodName: MARKET_METHOD_NAMES.BUY,
+    gas: GAS_CONSTANTS.MAX_GAS,
+    deposit: price.amount,
```
tifrel, line 28: Only applies for non-FT buy

packages/sdk/src/mb-impl/market.ts
```diff
@@ -0,0 +1,67 @@
...
+): TransactionArgs & TransactionAttachments => {
+  const { marketAddress, listAmount = 1 } = args;
+
+  const deposit = (0.01 * listAmount).toString();
+  return {
+    contractAddress: marketAddress,
+    args: {},
+    methodName: MARKET_METHOD_NAMES.DEPOSIT_STORAGE,
+    deposit: utils.format.parseNearAmount(deposit),
+    gas: GAS_CONSTANTS.MAX_GAS,
```
tifrel, line 65: I am pretty sure this can use default gas

cif:
Attempting to stream lerna output to debug build

packages/sdk/src/mb-impl/token.ts
```diff
@@ -0,0 +1,214 @@
...
+const getFactoryContract = (network: string): AccountId => {
+  if (network === Network.MAINNET) return MB_MAINNET_TOKEN_FACTORY;
+  else return MB_TESTNET_TOKEN_FACTORY;
+};
+
+// TODO: figure out a way to generate gas and deposit for each
+
+export const transfer = (args: TransferArgs): TransactionArgs => {
+  const { nftContractId, transfer } = args;
```
tifrel, line 33: If the agrs to `transfer` have a `transfer` field, that a) feels odd to me, and b) seems to be a very different API design than the `mint` function

packages/sdk/src/mb-impl/token.ts
```diff
@@ -0,0 +1,214 @@
...
+  } = args;
+
+  const { symbol, icon, baseUri, reference, referenceHash } = metadata;
+
+  const data = {
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    owner_id: ownerId,
+    metadata: {
+      spec: 'nft-1.0.0',
+      name: name.replace(/[^a-z0-9]+/gim, '').toLowerCase(),
```
tifrel, line 98: Shouldn't we return an error instead of silently overwriting user input? I feel this would be very annoying to debug.

packages/sdk/src/mb-impl/token.ts
```diff
@@ -0,0 +1,214 @@
...
+      symbol: symbol.replace(/[^a-z0-9]+/gim, '').toLowerCase(),
```
tifrel, line 99: Same here about returning an error instead of overwriting. Also currently a lot of symbols are all upper case, so why force them to be lower case now?

packages/sdk/src/mb-impl/token.ts
```diff
@@ -0,0 +1,214 @@
...
+      icon: icon ?? DEFAULT_MB_LOGO,
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      base_uri: baseUri ?? null,
```
tifrel, line 102: Might be worth a design consideration. With all the problems we had I am actually in favor of not having base URIs at all and just using full URIs in the token reference. This would probably make it easier to build IPFS minters (and whatever other minters we would like to have)

packages/sdk/src/mb-impl/token.ts
```diff
@@ -0,0 +1,214 @@
...
+      reference: reference ?? null,
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      reference_hash: referenceHash ?? null,
```
tifrel, line 105: So we are still not enforcing the proper hashing of the reference material?

packages/sdk/src/mb-impl/token.ts
```diff
@@ -0,0 +1,214 @@
...
+export const addMinter = (args: AddRemoveMinterArgs): TransactionArgs => {
+  const { minterId, nftContractId } = args;
+
+  return {
+    contractAddress: nftContractId,
+    args: {
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      account_id: minterId,
+    },
+    methodName: METHOD_NAMES.ADD_MINTER,
```
tifrel, line 158: Shouldn't we just use `batch_change_minters` for all cases so we can deprecate the old methods?

packages/sdk/src/mb-impl/token.ts
```diff
@@ -0,0 +1,214 @@
...
+    contractAddress: nftContractId,
+    args: {
+      grant: addMinters.length > 0 ? addMinters : undefined,
+      revoke: removeMinters.length > 0 ? removeMinters : undefined,
+    },
+    methodName: METHOD_NAMES.BATCH_CHANGE_MINTERS,
+  };
+};
+
+export const revoke = (args: RevokeAccountArgs): TransactionArgs => {
```
tifrel, line 190: Would name it `revokeApprovals`, I thought "Why do we need another method to revoke minters"?

Also, shouldn't there be a `approve` for symmetry? `buy` could reuse that fn

packages/react/src/hooks/useMinter.ts
cif: I like execute personally.  I had thought to make a separate @mintbase-js/rpc package for view calls but maybe we can keep in SDK

packages/sdk/src/mb-impl/market.ts
microchipgnu: Are we? Do you have docs for this? I have followed the ones I made two months ago.

packages/sdk/src/mb-impl/market.ts
microchipgnu: Oh naming was incorrect, it was actually the default. Thanks for pointing out.

packages/sdk/src/mb-impl/market.ts
microchipgnu: Right, can you point me out the right interface for FT buys?

packages/sdk/src/mb-impl/constants.ts
microchipgnu: Updated these.

packages/react/src/hooks/useMinter.ts
microchipgnu: Created a generic `execute` method that takes objects of the type `NearContractCall` or an array of `NearContractCall`

packages/sdk/src/mb-impl/token.ts
microchipgnu: Yes, if that is your plan on the contract level, let me know.

packages/sdk/src/mb-impl/token.ts
```diff
@@ -0,0 +1,214 @@
...
+export const removeMinter = (args: AddRemoveMinterArgs): TransactionArgs => {
+  const { minterId, nftContractId } = args;
+
+  return {
+    contractAddress: nftContractId,
+    args: {
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      account_id: minterId,
+    },
+    methodName: METHOD_NAMES.REMOVE_MINTER,
```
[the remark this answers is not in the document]
microchipgnu: Yes, if that is your plan on the contract level, let me know.

packages/sdk/src/mb-impl/token.ts
microchipgnu: Perfect!

packages/sdk/src/mb-impl/market.ts
tifrel: This will be done using `ft_transfer_call` on the FT contract: https://github.com/Mintbase/interop-market/blob/main/tests/__tests__/ft-offers.ava.ts#L253

PR #171, opened 2022-11-16
🐛 when allowance undefined due to nearjs api, set to default value

rubenmarcus:


cif (approved):

src/wallet.ts
```diff
@@ -316,16 +316,20 @@ export class Wallet {
 
     const { data: accessKey } = await this.viewAccessKey(accountId, publicKey)
 
-    const allowance = utils.format.formatNearAmount(
+    let allowance = utils.format.formatNearAmount(
       accessKey.permission.FunctionCall.allowance
     )
 
+    if (typeof allowance == 'undefined') {
+      allowance = '0.25'
```
sainthiago, line 324: would add this to a const (if it doesn't exist already)

rubenmarcus: done

sainthiago (approved):

PR #176, opened 2022-11-17
rollback release

rubenmarcus:
rollback version to previous that was set manually + rollback action files (that stopped working)

### DIFF
--- a/.github/workflows/new-release.yml
+++ b/.github/workflows/new-release.yml
@@ -1,14 +1,26 @@
 name:  Telegram new-release notification
 
+#on:
+#  push:
+#    branches:
+#      - main # Change this to your default branch
+
 on:
   release:
     types: [released]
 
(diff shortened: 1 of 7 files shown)


.github/workflows/new-release.yml
```diff
@@ -1,14 +1,26 @@
...
 jobs:
+  # This workflow contains a single job called "build"
   telegram-msg:
+    # The type of runner that the job will run on
     runs-on: ubuntu-latest
 
+    # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
```
cif, line 21: Do we need to even check out source code when all we do is notify telegram?

sainthiago (approved):

PR #180, opened 2022-11-18
♻️ 📄 Transfer as example

cif:
Possible [documentation](https://docs.mintbase.io/~/revisions/55ijAWqTOZJIySbbIi2T/dev/mintbase-sdk-ref/sdk/transfer) structure. 

Adding more tests (see react app) as well, but note that the packages/testing transfer was already completed prior to this PR.  When we work through the remaining methods, we should try to include at least an integration test if not an integration test and a react test (browser based)


cif:
Coveralls failed mergecheck on a friday night?  Anyway, we should probably merge the check into one push pipeline, docs are still good separate - it takes as little as 10 seconds sometimes to process the gitbooks readmes push.

packages/app/components/TransferTest.tsx
```diff
@@ -0,0 +1,60 @@
+import { ReactElement, useState, useEffect, SetStateAction } from 'react';
```
cif, line 1: Added react app test for transfer as well, can be repeated for all methods (optional, is slow)

PR #181, opened 2022-11-21
tokenById + useTokenById

rubenmarcus:


packages/data/src/api/tokenById/tokenById.ts
```diff
@@ -0,0 +1,47 @@
+import { fetchGraphQl } from "../../graphql/fetch";
+import { errorContractAddress, errorToken } from "./tokenById.mock";
+import { tokenByIdQuery } from "./tokenbyId.query";
+
+import { TokenByIdResults } from "./tokenById.types";
+
+export const checkParams = (tokenId: string, contractAddress: string) => {
+  if (typeof tokenId !== "string") {
+    return errorToken;
```
cif, line 9: Don't think we want to import anything from mocks into the implementation.  Not only is it confusing, I think this would include the big mock object in any bundle using this method

packages/data/src/api/tokenById/tokenById.ts
```diff
@@ -0,0 +1,47 @@
...
+  } else if (typeof contractAddress !== "string") {
```
cif, line 10: I think typescript should mostly help out here with the types and honestly, even if they were numeric graphql would figure this out.  Maybe we can just check for undefined and include message that like `did you forget to include contract address?`

packages/data/src/api/tokenById/tokenById.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+};
+
+export const tokenById = async (
+  tokenId: string,
+  contractAddress: string
+): Promise<TokenByIdResults> => {
+  const { valid, message } = checkParams(tokenId, contractAddress);
+
+  if (valid) {
```
cif, line 23: Personal preference perhaps, but to me it's much clearer to return early with validation errors rather than wrap the entire rest of the method e..g. `if (!valid) { tell the user and return }`

packages/data/src/api/tokenById/tokenById.types.ts
```diff
@@ -0,0 +1,55 @@
+export interface NftTokenData {
```
cif, line 1: Is this redundant to the generated graphql options?

packages/data/src/api/tokenById/tokenById.ts
rubenmarcus: ok,  will fix this by tomorrow

packages/data/src/api/tokenById/tokenById.types.ts
rubenmarcus: no the GraphQl generated is wrong it has props like listings, that if you put on the query will break it

packages/data/src/api/tokenById/tokenById.ts
rubenmarcus: ok,  will fix this by tomorrow

packages/data/src/api/tokenById/tokenById.test.ts
```diff
@@ -0,0 +1,98 @@
+import { tokenById, checkParams } from "./tokenById";
+import { TokenByIdResults } from "./tokenById.types";
+
+import {
+  errorContractAddress,
+  errorToken,
+  tokenByIdMock,
+} from "./tokenById.mock";
```
cif, line 8: Interesting linter didn't fix quotes, it should lint tests as well will double check that scripting.  I don't have huge opinions either way on it, but went with [single quotes](https://github.com/Mintbase/mintbase-js/blob/modular-refactor/.eslintrc.json#L42 )

packages/data/src/api/tokenById/tokenById.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    });
+
+    const contractAddress = 1;
+    const tokenId = "test.mintbase1.near";
+
+    jest.spyOn(methods, "checkParams").mockReturnValue(errorContractAddress);
+
+    const consoleSpy = jest.spyOn(console, "error");
+
+    // @ts-ignore
```
cif, line 88: Why do we need to ignore ts errors here?  Looks like a straight forward call to me

rubenmarcus: it is calling tokenId with number ( wrong type)

packages/react/src/hooks/useTokenById.ts
```diff
@@ -0,0 +1,44 @@
...
+  const [state, dispatch] = useReducer(reducer, initialState);
+
+  useEffect(() => {
+    let isCancelled = false;
+
+    if (state.isLoading) {
+      (async () => {
+        dispatch({ data: null, type: ActionTypes.REQUEST_STARTED });
+
+        const { data, error } = await tokenById(tokenId, contractAddress);
```
cif, line 21: State management is pretty heavy here for a lightweight hook that just fetches some data or an error, maybe look into something like [react query](https://tanstack.com/query/v4/docs/adapters/react-query) ? caveat, I have not looked into react query yet, but this feels over engineered

cif:
In regards to branch name, we're gonna be adding a lot more queries easier to track name like `add-token-by-id-query` vs `add-queries`

PR #182, opened 2022-11-22
tokenById + useTokenById

rubenmarcus:
- adds query tokenById
- adds hook useTokenById
- adds example on /app package
- adds integration Test

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203367697533179


sainthiago (approved):

cif:
Fixed up the publish action 🎉  seems to work well again.

This is starting to look great, adding documentation can be done in an subsequent PR if desired

packages/data/src/api/tokenById/tokenById.types.ts
```diff
@@ -0,0 +1,52 @@
+export interface NftTokenData {
+  animationUrl: string | null;
```
cif, line 2: If it's not super painful, might be nice to standardize these fields as `camelCase`. Not sure if we added that one to the lint rules

rubenmarcus: for that we would have to do a mapper , or change on the query itself , and change the fields coming from the backend

PR #188, opened 2022-11-25
Adding `rpc` module with block height fetching method

tifrel:


cif:
The require return type on type script is debatable, making linting easier:  install eslint plugin vscode then add to settings json:

"editor.codeActionsOnSave": {
    "source.fixAll.eslint": true
  },

packages/rpc/README.md
```diff
@@ -0,0 +1,6 @@
+[//]: # `{ "title": "@mintbase-js/rpc", "order": 1 }`
```
cif, line 1: The order here is how gitbook presents the menu, probably want to sort this closer to testing I think 6 or 7.  You can preview how the docs look by checking this PR https://github.com/Mintbase/gitbook-docs/pull/7

sainthiago (approved):

packages/rpc/README.md
tifrel: Done

PR #190, opened 2022-11-25
Adding txn status check to RPC module

tifrel:


packages/rpc/src/methods/txnstatus.ts
```diff
@@ -0,0 +1,27 @@
...
+    id: 'dontcare',
+    method: 'tx',
+    params: [txnHash, senderId],
+  });
+  if (res.error) throw res.error;
+
+  let failure = false;
+  let pending = false;
+  res.result.receipts_outcome.forEach((outcome: any) => {
+    console.log(outcome);
```
cif, line 20: 🧹

packages/rpc/src/util.ts
```diff
@@ -0,0 +1,13 @@
+import { NEAR_RPC_URL } from "./constants";
+import superagent from "superagent";
+
+export const requestFromNearRpc = async (
+  body: Record<string, any>
+): Promise<Record<string, any>> => {
+  const res = await superagent
```
rubenmarcus, line 7: is superagent really needed ? I mean pure fetch wont work?

packages/rpc/src/util.ts
```diff
@@ -0,0 +1,13 @@
...
+    .post(NEAR_RPC_URL)
+    .send(body)
+    .set("Content-type", "application/json");
+
+  return res.body;
```
rubenmarcus, line 12: and if res fail?

packages/rpc/src/methods/blockheight.ts
```diff
@@ -1,15 +1,11 @@
...
-    })
-    .set('Content-type', 'application/json');
-  return res.body.result.sync_info.latest_block_height;
+  const res = await requestFromNearRpc({
+    jsonrpc: "2.0",
+    id: "dontcare",
+    method: "status",
+    params: [],
+  });
+  return res.result.sync_info.latest_block_height;
```
rubenmarcus, line 10: and if res is empty , null or undefined?

packages/rpc/src/methods/txnstatus.ts
```diff
@@ -0,0 +1,27 @@
...
+    jsonrpc: "2.0",
+    id: "dontcare",
+    method: "tx",
+    params: [txnHash, senderId],
+  });
+  if (res.error) throw res.error;
+
+  let failure = false;
+  let pending = false;
+  res.result.receipts_outcome.forEach((outcome: any) => {
```
rubenmarcus, line 19: is there a situation where result  is undefined and result.receipts_outcome are not an array?

rubenmarcus (changes requested):
just some questions before approve it, but it seems ok

packages/rpc/src/util.ts
tifrel: Pure fetch won't work inside NodeJS

rubenmarcus (approved):
LGTM, just fix the conflicts

sainthiago (approved):

PR #194, opened 2022-11-28
Storage module

cif:


.github/workflows/publish.yml
```diff
@@ -26,7 +25,7 @@ jobs:
 
       - name: Build & Test
         run: |
-          npm ci
+          npm i
```
tifrel, line 28: But shouldn't it be `ci`?

packages/storage/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "files": [
+    "lib"
+  ],
+  "keywords": [],
+  "author": "",
+  "license": "MIT",
+  "dependencies": {
+    "@bundlr-network/client": "^0.7.7",
+    "near-api-js": "^0.44.2",
+    "superagent": "^8.0.3"
```
tifrel, line 21: According to Ruben, we are using isomorphic-unfetch as our HTTP request polyfill

packages/storage/src/uploads.ts
```diff
@@ -0,0 +1,72 @@
...
+  } catch (err: unknown) {
+    const httpError = err as HttpError;
+    console.error(
+      `Uploading file to arweave failed: ${httpError.status} ${httpError.response.text}`,
+    );
+    throw err;
+  }
+};
+
+// export const uploadBuffer = async (
```
tifrel, line 59: Dead code?

SurgeCode (approved):

secret.config.js
```diff
@@ -0,0 +1 @@
+process.env.MINTBASE_API_KEY=e886efa80e627f165ebbc7b4d942c172
```
tifrel, line 1: Should this be public?

tifrel (commented):
Changes in `/package-lock.json` are massive. Also changed `packages/rpc/package-lock.json` with no apparent reason. Shouldn't the latter already exist on `alpha`?

packages/storage/src/uploads.ts
cif: This is front @microchipgnu first pass at bring your own key / client side Bundlr direct vs the upload service

secret.config.js
cif: This key itself isn't sensitive but may be in the future, seems like this PR might now be missing commits :/

.github/workflows/publish.yml
cif: Thought changed default version in nvm but guess that doesn't work at system level.  Either one here is fine really, when we were dealing with the node version issues this was attempt to workaround

secret.config.js
cif: Ah, nm we are just using .env via dotconfig now

PR #201, opened 2022-11-30
ownedNftsByStores missing param

rubenmarcus:


SurgeCode (approved):
lgtm

packages/data/src/api/ownedNftsByStore/ownedNftsByStore.ts
```diff
@@ -20,9 +20,9 @@ export const ownedNftsByStore = async (
   const { data, error } = await fetchGraphQl<OwnedNftsData>({
     query: ownedNftsByStoreQuery,
     variables: {
-      ownerId,
-      contractAddress,
-      limit: pagination.limit,
+      accountId: ownerId,
```
cif, line 23: Is this just an update or do we want to address any differences in the docs?

rubenmarcus: this is just an update to make it work

rubenmarcus: @cif  because at the docs we ask the user to pass ownerId, but I previous added ownerId as the param name to the variables of the query, but the right param is accountId.

sainthiago (approved):

PR #202, opened 2022-11-30
feat(sdk): ✨Mint method restructure + docs + tests

SurgeCode:


packages/sdk/src/mint/mint.ts
```diff
@@ -0,0 +1,78 @@
...
+    console.error('There must be at least 2 accounts in splits');
+  }
+
+  if (amount && amount > 99) {
+    console.error('It is not possible to mint more than 99 copies of this token using this method');
+  }
+
+  if (royaltyPercentage && royaltyPercentage < 0 || royaltyPercentage > 0.5) {
+    console.error('Invalid royalty percentage');
+  }
```
SurgeCode, line 49: Thoughts on accepting 0 - 0.5 instead 0 - 5000? Seemed more intuitive to me

cif: Yeah, most humans aren't thinking about percentages in terms of basis points, that is a fintech thing and a source of confusion in terms of basic UI. @nategeier @tifrel might have some input

packages/sdk/src/mint/mint.ts
```diff
@@ -0,0 +1,78 @@
...
+  args: MintArgs,
+): NearContractCall => {
+  const { nftContractId, reference, ownerId, options = {}  } = args;
+
+  const { splits, amount, royaltyPercentage } = options;
+  //0.5 -> 5000
+  adjustSplitsForContract(splits);
+
+  if (splits && Object.keys(splits).length > 50) {
+    console.error('Splits cannnot have more than 50 entries');
```
cif, line 36: Should we be throwing these errors vs just logging them?  In cases where minting is going to fail for the end user I would think we wan to stop bad args from making it to NEAR at all

packages/sdk/src/mint/mint.test.ts
```diff
@@ -0,0 +1,178 @@
...
+  };
+
+  const optionsWithInvalidAmount = {
+    splits: { 'test': 2 },
+    amount: 300,
+    royaltyPercentage: 0.5,
+  };
+
+  const optionsWithInvalidSplits: MintOptions =  {
+    splits:{
```
cif, line 28: This is quite thorough, hard to fault that but would bigger intervals work? 5ths?

PR #208, opened 2022-12-05
add isWaitingforConnection as false on modal close

rubenmarcus:


sainthiago (approved):

packages/react/src/WalletContext.tsx
```diff
@@ -57,11 +57,21 @@ export const WalletContextProvider: React.FC<React.PropsWithChildren> = (
     });
   }, [setup]);
 
+  const onCloseModal = (): void => {
+    setIsWaitingForConnection(false);
+  };
+
+  if (typeof window !== 'undefined') {
+    const closeButton = document?.getElementsByClassName('close-button')[0];
```
cif, line 65: Not a bad idea to use the DOM directly like this, but would have to ensure the node is there before binding the event listener

cif: Wait, does this actually work?  I wonder if getting a re-render from state update 🤔  cool.

cif: Added a warning just in case :)

PR #212, opened 2022-12-07
fix(sdk): export both v1 and new sdk methods

SurgeCode:


packages/sdk/src/index.ts
```diff
@@ -1,18 +1,16 @@
...
+export * from './batchChangeMinters/batchChangeMinters';
+export * from './deployContract/deployContract';
+export * from './transferContractOwnership/transferContractOwnership';
+export * from './list/list';
+export * from './buy/buy';
+export * from './depositStorage/depositStorage';
+export * from './mint/mint';
+export * from './revoke/revoke';
+
+
```
tifrel, line 16: Don't you need to `export * from './v1'` here to export both? Should we namespace it?

SurgeCode: Long story but typescript makes it super hard
if we did `export * from './v1'` we would need to do 
`import {v1} from '@mintbase-js/sdk'`
`const {mint} = v1`

I just added a index js to the v1 folder and now we can do 

`import {v1} from '@mintbase-js/sdk/lib/v1'`

SurgeCode: Also it seems that on omnisite we aren't even using that many wrappers to the point where only 1 or 2 changes were needed and we wont even need to import v1 there at all

cif (approved):

PR #218, opened 2022-12-13
legacy listing deposit

rubenmarcus:


microchipgnu (approved):

packages/sdk/src/v1/constants.ts
```diff
@@ -27,8 +29,11 @@ export const GAS_CONSTANTS = {
 export const DEPOSIT_CONSTANTS = {
   ONE_YOCTO: '1',
   ZERO_YOCTO: '0',
+  LEGACY_LISTING: '0.0044',
```
tifrel, line 32: This is in NEAR, while the other two are in yoctoNEAR?

tifrel (changes requested):

PR #221, opened 2022-12-14
feat(sdk): Make execute composable and return all results

SurgeCode:
This pr adds support for the composition of various mb-js methods into a single call which can get executed transparently by the user as if it was a single function.
It also solves the problem of not getting status and results from batch transactions.
Integration/Unit tests were added to validate that

### DIFF
--- a/packages/sdk/src/execute.ts
+++ b/packages/sdk/src/execute.ts
@@ -29,11 +29,10 @@ const validateSigningOptions = ({ wallet, account }: NearCallSigningOptions): vo
 
 /**
  * Base method for executing contract calls.
- *
- * @param call an array or single instance of {@link NearContractCall} to execute
  * @param signing object containing either near wallet selector
+ * @param calls  {@link NearContractCall[]} any number of of single calls or compositions
  *  wallet: {@link Wallet} or account: {@link Account}, defaults to wallet when present
- * @returns a result for single transactions of {@link FinalExecutionOutcome}, or void for batches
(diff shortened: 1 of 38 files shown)


packages/sdk/src/execute.ts
```diff
@@ -41,33 +36,68 @@ const validateSigningOptions = ({ wallet, account }: NearCallSigningOptions): vo
...
-    return batchExecuteWithNearAccount(call, account);
+// account call translation wrappers https://docs.near.org/tools/near-api-js/faq#how-to-send-batch-transactions
+// TODO: share batch signature with wallet selector sendAndSignTransaction when method becomes public
+
+const batchExecuteWithNearAccount = async (
+  calls: ContractCall[],
+  account: Account,
+): Promise<FinalExecutionOutcome[]> => {
+  const outcomes: FinalExecutionOutcome[] =[];
+  for (const call of calls) {
```
cif, line 68: In the future we might consider a parallel execution option, but for our only "real world" batch use case (storage deposit > listing) we actually want series

cif (commented):
Overall looks great, should we update the README at all with API changes

packages/sdk/src/execute.ts
SurgeCode: I added the information that they are executed in series to the documentation. Let's for sure add the parallel execution in the future.

sainthiago (approved):

Issue #222, opened 2022-12-14
uploadFileToArweave does not work with Buffer


elliotBraem:
I'm uploading metadata to form as the reference for a mint. For example:

```
const metadata = Buffer.from(JSON.stringify({
    extra: "sample"
  }));
const response = await uploadFileToArweave(metadata, "sample-name");
const url = `https://arweave.net/${response.id}`;
```

If I run this, I will get the following exception, which is misleading because the actual exception being thrown does not work with httpError.response.text:
```
Cannot read properties of undefined (reading 'text')
```

The actual exception is:
```
Failed to execute 'append' on 'FormData': parameter 2 is not of type 'Blob'.
```

I think somebody just needs to wrap the buffer in "new Blob([buffer])". Wasn't caught because append function is mocked in test.

cif:
I now recall quickly authoring this function and we actually aren't using it anywhere yet.   As you point out, this could use an integration test.

Would you mind confirming the environment you're running this in, including node version if server side?   

I'm looking at the integration tests run against the service which are passing and don't wrap the buffer in Blob, so I wonder if env changes how superagent/supertest is treating the bytes under the hood 🤔

elliotBraem:
Oh interesting; I’m using this client side and this happened in development, but the same error is happening in my production build as well. 

This is all I’m doing:
```
export async function mintThing(thingId: string, args: CreateThingBlockchainArgs): Promise<void | FinalExecutionOutcome> {
  if (!isUuid(thingId)) {
    throw new Error(NOT_VALID_UUID);
  }

  const metadata = Buffer.from(JSON.stringify({
    extra: thingId
  }));
  const response = await uploadFileToArweave(metadata, thingId);
```

Creating an object with a single field, making it a JSON, turning into Buffer, passing to function.
Passing File has been working for me, Buffer is not.

cif:
The integration tests are all run in node, so I am guessing this has to do with the browser env.  Someday, we hope to have test automation in the packages/app browser test suite, but until then we can do a manual test there and be sure this works with client side JSON.  Have you attempted to upload from `File` as well?  That is also on the list, so maybe we can do it in a single PR

cif:
@elliotBraem looks like @microchipgnu has created a draft with starting point for the browser based upload function: https://github.com/Mintbase/mintbase-js/pull/223 

We are short staffed this week, but will work on some final testing and documentation and get this merged ASAP.

[...]

PR #223, opened 2022-12-16
fix(storage): upload works in browser

microchipgnu:


SurgeCode:
There's an integration test for uploading in the testing package that we didn't complete if you feel like it, not urgent but if you're feeling it

cif (commented):
Let's at a minimum, document these methods in README with examples. Even better, create an example in packages/app that does the file upload and use that working code as the example in docs

microchipgnu:
@cif @SurgeCode added documentation and an example.

packages/storage/src/README.md
```diff
@@ -0,0 +1,45 @@
+[//]: # `{ "title": "upload", "order": 0.1 }`
+
+# Upload in the Browser
+
+It's possible to use [File](https://developer.mozilla.org/en-US/docs/Web/API/File) to upload a file. 
+
+
+Example usage of storage method in a hypothetical React component:
+{% code title="StorageUI.ts" overflow="wrap" lineNumbers="true" %}
+
```
SurgeCode, line 10: Maybe adding a bit of a description here of what the method is doing and what it returns would be cool for people who don't know what is happening.

microchipgnu: That's good will do that.

sainthiago (approved):

SurgeCode (approved):
Looks good! 👍

cif:
This should address and close https://github.com/Mintbase/mintbase-js/issues/222

PR #226, opened 2022-12-22
🩹 Wallet callback

SurgeCode:


packages/sdk/src/execute.ts
```diff
@@ -71,6 +71,7 @@ const batchExecuteWithNearAccount = async (
         contractId: call.contractAddress,
         methodName: call.methodName,
         args: call.args,
+        walletCallbackUrl: call.callbackUrl,
```
cif, line 74: Since this method is currently specific to near-api-js vs wallet selector we won't need callback here.  If you look at the comments above this method though, it does seem like at some point we could make the signatures match better if not perfectly.  

The signing is all near-api-js under the hood anyway, seems like the variance is in how browser/extension based wallets handle the outcomes

PR #235, opened 2022-12-28
Fix(storage): upload browser fix 

SurgeCode:


packages/storage/src/uploads.ts
```diff
@@ -66,7 +60,7 @@ export const uploadFile = async (
   file: File,
 ): Promise<ArweaveResponse> => {
   if (MINTBASE_API_KEY == MINTBASE_API_ANON_USER) {
-    console.warn(ANON_USER_WARNING);
+    console.warn('Warning: you are using the anonymous mintbase api key. You need to specify yours in order to upload a file.');
```
cif, line 63: Let's leave this warning, but say "in the future, you will need to obtain a session key from mintbase connect"

SurgeCode: Good idea

sainthiago (approved):

PR #238, opened 2022-12-29
fix sdk react examples

rubenmarcus:
It fixes some of the examples (types, component names, handle function names) as also add some more that was missing


sainthiago (approved):

SurgeCode (approved):
Good job, nice cleanup ✅

PR #244, opened 2023-01-04
Add get token owner query

sainthiago:
This PR adds Get Token Owner Query to the @data package


microchipgnu (changes requested):
Instead of forcing "Format document", just enable format on save to use our ESLint config.

SurgeCode (approved):

rubenmarcus (approved):

PR #248, opened 2023-01-04
Data token provenance query

sainthiago:


packages/data/src/api/tokenProvenance/tokenProvenance.ts
```diff
@@ -0,0 +1,44 @@
...
+      return { error: errorToken.message };
+    }
+  }
+
+  const { data, error } = await fetchGraphQl<TokenProvenanceData>({
+    query: getTokenProvenance,
+    variables: {
+      tokenId,
+      contractAddress,
+      limit: pagination?.limit ?? 12,
```
cif, line 36: Wonder if we should make the 12 a const like `DEFAULT_RECORDS_PER_PAGE` and add this to other data query methods

cif (approved):

PR #252, opened 2023-01-06
Requiring media to mint

tifrel:
I guess this breaks omni-site. We decided a while back that on-chain media would be very desirable, as the MB way of having media in the reference blob is supported by standards, but we are about the only folks in the ecosystem doing it that way, and it's highly likely that it will cause problems with third parties wanting to integrate Mintbase.

### DIFF
--- a/packages/sdk/src/mint/mint.ts
+++ b/packages/sdk/src/mint/mint.ts
@@ -6,6 +6,7 @@ import { NearContractCall } from '../execute';
 export type MintArgs =  {
   nftContractId?: string;
   reference: string;
+  media: string;
   ownerId: string;
   options?: MintOptions;
 };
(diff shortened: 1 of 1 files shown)


rubenmarcus (approved):

packages/sdk/src/mint/mint.ts
```diff
@@ -61,6 +62,7 @@ export const mint = (
       owner_id: ownerId,
       metadata: {
         reference: reference,
+        media: media,
```
cif, line 65: Thoughts on extending this capability to other/all on chain standardized NEP 171 options?

cif:
Created a task to pick this up and carry it far as we think makes sense https://app.asana.com/0/1203642570392111/1203702048447301

PR #256, opened 2023-01-09
Add Owned Stores by OwnerId method in data package

sainthiago:


SurgeCode (approved):
Seems good, only thing i noticed was there wasn't limit and offset but i don't think people will ever have enough stores for that to be relevant

PR #259, opened 2023-01-10
Data near price method

sainthiago:


cif (approved):

package.json
```diff
@@ -85,6 +85,7 @@
     "packages/*"
   ],
   "dependencies": {
+    "fetch-mock": "^9.11.0",
     "jest-fetch-mock": "^3.0.3"
```
rubenmarcus, line 89: No need for this

rubenmarcus (changes requested):

packages/data/src/api/nearPrice/nearPrice.ts
```diff
@@ -0,0 +1,41 @@
...
+    );
+
+    const res: BinanceNearPriceData = await req.json();
+
+    if (!res.price) {
+      const price = await nearPriceFallback();
+      return { nearPrice: price.near.usd };
+    }
+
+    return { nearPrice: res.price };
```
rubenmarcus, line 28: if you return as it is now on the nearPrice function you will need to return price.nearPrice
```suggestion
    return res.price ;
```

packages/data/src/api/nearPrice/nearPrice.ts
```diff
@@ -0,0 +1,41 @@
...
+  try {
+    const req = await fetch(
+      BINANCE_API,
+    );
+
+    const res: BinanceNearPriceData = await req.json();
+
+    if (!res.price) {
+      const price = await nearPriceFallback();
+      return { nearPrice: price.near.usd };
```
rubenmarcus, line 25: return price.near.usd
```suggestion
      return price.near.usd;
```

rubenmarcus (changes requested):

rubenmarcus (approved):

PR #266, opened 2023-01-16
Add Loading flag in data package

sainthiago:


packages/data/src/utils.ts
```diff
@@ -1,7 +1,12 @@
 import { ParsedDataReturn } from './types';
 
 
-export const parseData = <T>(data: T, error: null | string, errorMsg: string): ParsedDataReturn<T> => {
+export const parseData = <T>(data: T, error: null | string, errorMsg: string, loading?: boolean): ParsedDataReturn<T> => {
+  
+  if (loading) {
```
cif, line 6: If we really wanted to abstract this into the fetching lib itself, we would have to use generators or perhaps observables/subscription to yield loading state, then the finished/error state when the process completes as-is, this will return loading state and not finish afterword.  Not opposed but we definitely want to think this one through

cif:
As with comment on that line, we would want to think through how we handle stateful loading before we proceed here.  We could also elect not to do this with hooks in the more library/specific code and document there, or let folks manage that state themselves

PR #271, opened 2023-01-17
Add metadataByMetadaId method in data package

sainthiago:


packages/data/src/api/metadataByMetadataId/metadataByMetadataId.test.ts
```diff
@@ -0,0 +1,35 @@
...
+      // console.log('Suppressed console error.');
+    });
+  });
+
+  it('returns data', async () => {
+    (GraphQLClient as jest.Mock).mockImplementationOnce(() => ({
+      request: (): Promise<MetadataByMetadataIdQueryResult> =>
+        Promise.resolve(metadataByMetadataIdMock),
+    }));
+    const result = await metadataByMetadataId('test' );
```
rubenmarcus: const result = await metadataByMetadataId('test' );
```suggestion
    const result = await metadataByMetadataId('test');
```

rubenmarcus (approved):

PR #272, opened 2023-01-17
feat: Enable Here wallet

cif:


sainthiago (approved):

rubenmarcus (approved):

packages/sdk/src/config/README.md
```diff
@@ -5,72 +5,69 @@ This is the method to setup the global variables object on all your mintbase-js
...
+  //can retrieve the keys in any part of your application.
+  console.log(mbjs.keys , 'global keys of all mintbase-js packages')
+```
 
-  ```typescript
-    import { tokenById } from '@mintbase-js/data';
 
-    const { data, error } = await tokenById('5','logout.mintbase1.near', 'mainnet');
-  ```
+### 2. Pass the network directly to methods (where applicable).
```
rubenmarcus, line 29: contractAddress is one env variable that also is supported in some methods ( and also can be set on mbjs.config )

rubenmarcus (changes requested):

packages/sdk/src/config/README.md
cif: Let's address some of this in the testing task we have in the sprint. Restore integration tests as well as get some unit testing around using keys inside of the methods. I just had to fix a bug while testing HERE wallet

cif: Can you un-request these changes and we can add in a second pass? Just trying to ship the HERE wallet now

rubenmarcus (approved):

PR #274, opened 2023-01-18
Minting with metadata

tifrel:


sainthiago (approved):

packages/sdk/src/mint/mint.ts
```diff
@@ -5,11 +5,28 @@ import { NearContractCall } from '../execute';
 
 export type MintArgs =  {
   nftContractId?: string;
-  reference: string;
   ownerId: string;
+  metadata: TokenMetadata;
```
cif: This is a breaking API change, perhaps we can still accept reference, then make an additional property in `options` called `additionalMetadata` , then in args: `metadata: { reference, ...additionalMetadata }`

tifrel: Did we decide not to do any breaking changes? Shouldn't typing + versioning be enough?

cif: Fair point, when you're in alpha/beta anything server `0.x` technically all bets are off.  Maybe we release beta this week with the clear message "you should pin versions and look for changes to this API each time you update"

packages/sdk/src/mint/mint.ts
```diff
@@ -26,14 +43,28 @@ export type Splits = Record<string, number>;
...
 
   const { splits, amount, royaltyPercentage } = options;
-  
-  if (nftContractId == null) {
+
+  if (nftContractId === null) {
     throw new Error('You must provide a nftContractId or define a NFT_CONTRACT_ID environment variable to default to');
   }
 
+  if (!noReference && !metadata.reference) {
```
cif, line 30: Very nice

cif (approved):

packages/sdk/src/mint/mint.ts
```diff
@@ -26,14 +43,28 @@ export type Splits = Record<string, number>;
...
+    throw new Error('You must provide reference in your metadata or explicitly opt out of using reference');
```
cif: Looks like the conflict here moved these messages into constants which is nice, be good to follow that pattern as well

tifrel: Done

packages/sdk/src/mint/mint.ts
tifrel: I left it in flexibly now, but I prefer the `metadata.reference` API. Requires some logic to check if it's conflicting or not. If we prioritize API ergonomics and clarity, let's remove it.

cif: The current version I pulled still had the breaking change, removed the conflict message and updated the docs.  I think we will just roll the breaking change as part of beta release task today

PR #277, opened 2023-01-19
feat: Add method to resolve payouts from RPC for UI

tifrel:


packages/rpc/src/util.ts
```diff
@@ -1,16 +1,50 @@
...
+    params: {
+      request_type: "call_function",
+      finality: "final",
+      account_id: contractId,
+      method_name: method,
+      args_base64,
+    },
+  });
+
+  console.log(res);
```
cif: Github is breaking my brain, this is not on the local branch when I pull 🤔

PR #279, opened 2023-01-19
 Fix callback url in all wallets

rubenmarcus:
## changes

**@mintbase-js/sdk**

- add support to the user set callbackUrl + callbackArgs on execute method
- gets callbackUrl from mbjs.config
- redirects to the page post transaction in all wallets ( Injected Wallets too ) 
- add type guard to return args of the sdk methods
- adds README for execute method
- some minor changes in other sdk methods README
- re-structure all sdk methods, separating the code by concern , types in types.ts file on root, utils in utils.ts file on root , or in the same folder file if too many utils ( ex: execute.utils.ts )
- adds extra unit test on execute method

**@mintbase-js/data**
- fixes some README issues on data package


packages/sdk/src/burn/burn.ts
```diff
@@ -1,23 +1,15 @@
...
-export const DEPOSIT_FOR_BURN = ONE_YOCTO;
-export const GAS_FOR_BURN = GAS;
 
 /**
  * Burns one or more tokens from a given contract.
  * @param burnArguments {@link BurnArgs}
  * @returns contract call to be passed to @mintbase-js/sdk execute method
  */
-export const burn = ({ tokenIds, contractAddress = mbjs.keys.contractAddress }: BurnArgs): NearContractCall => {
+export const burn = ({ tokenIds, contractAddress = mbjs.keys.contractAddress }: BurnArgs): NearContractCall<BurnReturnArgs> => {
```
tifrel: Why are the other generics `<MethodArgs>`, but this `<MethodReturnArgs>`?

packages/sdk/src/execute/execute.utils.ts
```diff
@@ -0,0 +1,184 @@
+import type { Wallet, FinalExecutionOutcome } from '@near-wallet-selector/core';
```
tifrel, line 1: Wouldn't this be `execute/utils.ts` instead of `execute/execute.utils.ts`?

packages/sdk/src/list/README.md
```diff
@@ -30,11 +30,10 @@ export type ListArgs = {
   }
 ```
 
-
 ## React example
 
 
-Example usage of deployContract method in a hypothetical React component:
+Example usage of list method in a hypothetical React component:
```
tifrel, line 36: Nice catch :D

packages/sdk/src/utils.ts
```diff
@@ -0,0 +1,4 @@
+
+export function standardizeString(name: string): string {
+  return name.replace(/[^a-z0-9]+/gim, '').toLowerCase();
```
tifrel, line 3: Is this tested? This should just remove every char that is not alphanumeric, right?

tifrel (approved):
I find naming for the generic of `NearContractCall` a bit confusing, maybe we should have a pattern of `NearContractCall<MethodXCall>`. `MethodXArgs` would be the arguments needed for the call and `ReturnArgs` sounds a bit clunky and contradictory to me.

packages/sdk/src/burn/burn.ts
cif: Agree with feedback on "return args", as they are not really arguments to the function but a return "type", maybe something like `Response<Burn>` or `BurnResponse`, `BurnResult` etc

sainthiago (approved):

packages/sdk/src/execute/execute.utils.ts
rubenmarcus: this is to keep naming convention as the tests we add the name of the method before too like transfer.test.ts

packages/sdk/src/burn/burn.ts
rubenmarcus: as explained in telegram, the naming was due that this is the typing for the args property inside NearContractCall, thats why I added Args in the end

packages/sdk/src/utils.ts
rubenmarcus: this code is from @SurgeCode , I didnt test it

cif (approved):

PR #284, opened 2023-01-24
✨Reference uploads from new endpoint

SurgeCode:


packages/storage/src/uploads.ts
```diff
@@ -124,29 +125,53 @@ export const uploadFile = async (
...
+  });
 
-  if (canUploadAnimation) {
-    referenceObject[animation_url] = (await uploadFile(animation_url)).id;
-  }
+  try {
+    const request = await fetch(`${ARWEAVE_SERVICE_HOST}/reference`, {
+      method: 'POST',
+      headers: {
+        'mb-api-key': MINTBASE_API_KEY,
```
cif, line 151: About to be working on what we want to be sending in this header, but probably better to have this working first

cif (approved):

sainthiago (approved):

PR #290, opened 2023-01-27
fix(storage): Reference edge cases and testing

SurgeCode:


packages/storage/src/uploads.ts
```diff
@@ -180,3 +171,35 @@ export function getFileFromObject(referenceObject: unknown): File {
...
+    const canBeUploaded = value instanceof File && value.size < MAX_UPLOAD_BYTES; 
+    const invalidFile = !hasCorrectMediaType && (value instanceof File)
+    const mediaTypeWithoutFile = hasCorrectMediaType && (typeof(value) == "string")
+
+    if(invalidFile){
+      // example title: File 
+      throw new Error("The provided field has a key that is not recognized by our service and will not be uploaded to arweave, try using media, animation_url or document")
+    }
+
+    if(mediaTypeWithoutFile){
```
cif: super nit picky, spaces on these two if statements don't match the lower ones thought we had auto-save linting rule around those

SurgeCode: Fixed

cif (approved):

sainthiago (approved):

PR #294, opened 2023-02-01
🧼 Beta Release Cleanup

cif:
Preparing for beta release:

- [x] Additional pass at docs
- [x] Fix integration tests
- [X] Refactor config types
- [X] Misc linter rules (auto fix only)

https://app.asana.com/0/1203748907229589/1203856806301140


packages/sdk/src/types.ts
```diff
@@ -57,21 +57,23 @@ export enum RPC_ENDPOINTS  {
...
-  network: Network | string;
+export type ConfigOptions = {
+  network?: Network | string;
   contractAddress?: string;
   callbackUrl?: string;
+  apiKey?: string;
 }
 
-export interface MbJsConfigObj extends MBJS_CONFIG_PARAMS {
+export interface ConfigOptionsObj extends ConfigOptions {
```
cif, line 67: Generally kind of okay with `enum` being upper case... but I think we are using these as `const.  Unless we actually intend to make any of these items enumerable i.e. mappable via numeric index from database enum field, dropdown index or something can just keep them as const.  

May or may not address this here, if anyone loves enums strongly for other reasons I am not considering, please speak up.

packages/sdk/src/execute/execute.utils.ts
```diff
@@ -1,30 +1,33 @@
 import type { Wallet, FinalExecutionOutcome } from '@near-wallet-selector/core';
 import { BrowserWalletSignAndSendTransactionParams } from '@near-wallet-selector/core/lib/wallet';
 import type { providers, Account } from 'near-api-js';
-import { NearContractCall, CallBackArgs, ContractCall, TxnOptionalSignerId, NearExecuteOptions, ExecuteArgsResponse } from '../types';
+import { CallBackArgs, ContractCall, TxnOptionalSignerId, NearExecuteOptions, ExecuteArgsResponse } from '../types';
 import BN from 'bn.js';
 import { NoSigningMethodPassedError } from '../errors';
+import { ComposableCall } from '../types';
```
rubenmarcus: this here could be added on line 4 is same import

rubenmarcus (approved):
just one small change on import lgtm

packages/sdk/src/types.ts
cif: Going to leave this as is based on internal conversation

packages/data/src/unwrap.ts
```diff
@@ -0,0 +1,3 @@
+
+export { ownedTokensThrowOnError as ownedTokens } from './api/ownedTokens/ownedTokens';
+export { tokensByStatusThrowOnError as tokensByStatus } from './api/tokensByStatus/tokensByStatus';
```
cif, line 3: Just want to quickly call attention to this one- we may want to extend this to other methods depending on the application.  Generally being able to consume the `{  data, error }` is nice in client apps for error display, but the ergonomics around one that simply throws the error or returns data is nice for most applications.

cif: To be clear, supporting both is the long term goal, see the update in README as well

sainthiago (approved):

packages/data/src/unwrap.ts
rubenmarcus: the idea around `{data,error} `was to have same pattern as Apollo and other libs have

SurgeCode (approved):

PR #310, opened 2023-02-20
fix(sdk): underscore `_` in `CONTRACT_DEPOSIT` const causes error

microchipgnu:
Screen Shot 2023-02-20 at 4 29 25 PM


sainthiago (approved):

cif (approved):
If this isn't urgent to merge, we noticed there may possible be an issue with the minting deposit calculation underestimating could be good to add

microchipgnu:
@cif lets merge this then. This is preventing contract deployment from @mintbase-js/sdk. I found a way around that by overriding the deposit. Default values won't work.

PR #312, opened 2023-02-22
Simpler Minting With Royalties

tifrel:


packages/sdk/src/mint/mint.ts
```diff
@@ -16,14 +16,13 @@ export const mint = (
     contractAddress = mbjs.keys.contractAddress,
     metadata,
     ownerId,
-    options = {},
+    royalties,
+    amount = 1,
```
tifrel: Shouldn't `amount` and `tokenIdsToMint` be mutually exclusive arguments?

packages/sdk/src/mint/mint.ts
```diff
@@ -16,14 +16,13 @@ export const mint = (
     contractAddress = mbjs.keys.contractAddress,
     metadata,
     ownerId,
-    options = {},
+    royalties,
```
tifrel, line 19: No more splits? :(

packages/sdk/src/mint/mint.ts
```diff
@@ -36,75 +35,91 @@ export const mint = (
...
   }
 
   if (amount && amount > 125) {
     throw  new Error(ERROR_MESSAGES.MAX_AMOUT);
   }
-
-  if (royaltyPercentage && royaltyPercentage < 0 || royaltyPercentage > 0.5) {
-    throw new Error(ERROR_MESSAGES.INVALID_ROYALTY_PERCENTAGE);
+  if (tokenIdsToMint && tokenIdsToMint.length == 0) {
+    throw  new Error(ERROR_MESSAGES.MAX_AMOUT);
```
tifrel: Should be a different error msg I guess

packages/sdk/src/mint/mint.ts
```diff
@@ -36,75 +35,91 @@ export const mint = (
...
   }
 
+  const adjustedTokenCount = tokenIdsToMint?.length > amount? tokenIdsToMint.length : amount;
```
tifrel: This is more headaches. So we can mint 20, and maybe I only want to specify 5 token IDs, but it's still valid?

packages/sdk/src/mint/mint.ts
```diff
@@ -36,75 +35,91 @@ export const mint = (
...
 
-  if (splits) {
+  let adjustedRoyalties: Splits = {};
+  let royaltyTotal: number;
+
+  if (royalties) {
     // FIXME: suggest we use % (ints) vs float here
     // 0.5 -> 5000
-    adjustSplitsForContract(splits);
+    royaltyTotal = getRoyaltyTotal(royalties);
```
tifrel, line 44: Why iterate over royalties twice? Esp since we are calculating totals in both cases

tifrel: Ah, it's necessary for the adjustment

packages/sdk/src/mint/mint.ts
```diff
@@ -36,75 +35,91 @@ export const mint = (
...
-  Object.keys(splits).forEach(key => {
-    counter += splits[key];
-    splits[key] *= 10_000;
+  const result: Splits = {};
+  Object.keys(royalties).forEach(key => {
+    if (royalties[key] <= 0) {
+      throw new Error (ERROR_MESSAGES.NEGATIVE_ROYALTIES);
+    }
+    const adjustedAmount = (royalties[key]/ royaltyTotal) * 10000 ;
+    result[key] = adjustedAmount;
```
tifrel, line 116: Don't we need to `Math.round` somewhere?

packages/sdk/src/mint/README.md
```diff
@@ -4,15 +4,15 @@
...
 
-Royalties and splits can be configured to provide a customized flow of funds to up to 50 people as explained below.
+Royalties can be configured to provide a customized flow of funds as explained below.
 
 It is possible to configure the amount of copies you want to mint through the `amount` field, but currently they will all share the same reference material.
 
 The nftContactId can be supplied as an argument or through the `TOKEN_CONTRACT` environment variable.
 
-**As with all new SDK api methods, this call should be wrapped in [execute](../#execute) and passed a signing method**
+**As with all new SDK api methods, this call should be wrapped in [execute](../#execute) and passed a signing method. For a guide showing how to make a contract call with mintbasejs click [here](https://docs.mintbase.xyz/dev/getting-started/make-your-first-contract-call-deploycontract)**
```
tifrel: Do we always just call it mintbasejs? Or do we use e.g. MintbaseJS for public consumption?

packages/sdk/src/errorMessages.ts
```diff
@@ -3,16 +3,23 @@ const SUPPORT = '- further help available on our telegram channel: https://t.me/
 
 export const ERROR_MESSAGES = {
   CONTRACT_ADDRESS: `You must provide a contractAddress, define contractAddress on mbjs.config,or a CONTRACT_ADDRESS enviroment variable to default to ${DEFAULT_SUPPORT}`,
-  INVALID_ROYALTY_PERCENTAGE: `Invalid royalty percentage ${SUPPORT}`,
+  INVALID_ROYALTY_PERCENTAGE: `Invalid royalty percentage, it must be between 0 and 0.5 ${SUPPORT}`,
```
tifrel, line 6: Did we settle for 0-1 range instead of 0-100?

packages/sdk/src/mint/mint.test.ts
```diff
@@ -133,60 +57,144 @@ describe('mint method tests', () => {
...
         token_ids_to_mint: [123, 456],
       },
       deposit: '14150000000000000000000',
       gas: GAS,
     });
   });
 
-  test('mint with invalid percentage', () => {
-    expect(()=> mint({
+  test('mint with flaoting point royalties', () => {
```
tifrel: floating

packages/sdk/src/mint/mint.test.ts
```diff
@@ -133,60 +57,144 @@ describe('mint method tests', () => {
...
-          test2: 5000,
-        },
         token_ids_to_mint: [123, 456],
       },
       deposit: '14150000000000000000000',
       gas: GAS,
     });
   });
 
-  test('mint with invalid percentage', () => {
```
tifrel, line 150: Why remove all the edge cases? We should also make sure mbjs doesn't allow you to construct invalid transactions.

packages/sdk/src/mint/mint.test.ts
```diff
@@ -133,60 +57,144 @@ describe('mint method tests', () => {
...
       metadata: { reference, media },
       ownerId: ownerId,
-      options: optionsWithInvalidPercentage,
-    })).toThrow();
-  });
+      royalties: { test: 0.23, test1: 0.12654, test2: 0.04421 },
+      tokenIdsToMint: [123, 456],
+    });
 
-  test('mint with invalid amount', () => {
```
tifrel, line 159: edge case dropped?

packages/sdk/src/mint/mint.test.ts
```diff
@@ -133,60 +57,144 @@ describe('mint method tests', () => {
...
+          },
+        },
+        token_ids_to_mint: [123, 456],
+      },
+      deposit: '16550000000000000000000',
+      gas: GAS,
+    });
   });
 
-  test('mint with invalid splits', () => {
```
tifrel, line 168: Should check validity of royalties

packages/sdk/src/mint/mint.test.ts
```diff
@@ -133,60 +57,144 @@ describe('mint method tests', () => {
...
       metadata: { reference, media },
       ownerId: ownerId,
-      options: optionsWithInvalidSplits,
-    }))).toThrow();
-  });
+      royalties: { test: 0.23, test1: 0.12654, test2: 0.04421 },
+      amount: 2,
+    });
 
-  test('mint without reference', () => {
```
tifrel, line 177: edge case dropped?

packages/sdk/src/mint/mint.test.ts
```diff
@@ -133,60 +57,144 @@ describe('mint method tests', () => {
...
+          },
+        },
+        token_ids_to_mint: null,
+      },
+      deposit: '16550000000000000000000',
+      gas: GAS,
+    });
   });
 
-  test('mint without media', () => {
```
tifrel, line 185: edge case dropped?

packages/sdk/src/mint/mint.ts
```diff
@@ -36,80 +35,122 @@ export const mint = (
...
+  let adjustedRoyalties: Splits;
+  let royaltyTotal: number;
+  let roundedRoyalties: Splits;
 
-  if (splits && Object.keys(splits).length > 50) {
-    throw new Error(ERROR_MESSAGES.MAX_SPLITS);
+  //royalties adjustments to make devx better
+  if (royalties) {
+    royaltyTotal = getRoyaltyTotal(royalties);
+    adjustedRoyalties = adjustRoyaltiesForContract(royalties, royaltyTotal);
```
tifrel, line 45: Why adjust and round in two steps?

packages/sdk/src/mint/mint.ts
```diff
@@ -36,80 +35,122 @@ export const mint = (
...
 
-  if (splits && Object.keys(splits).length < 2) {
-    throw new Error(ERROR_MESSAGES.SPLITS);
+  //if royalties exist they need to be populated
+  if (royalties && Object.keys(royalties).length < 1) {
+    throw new Error(ERROR_MESSAGES.MIN_ROYALTIES);
   }
 
   if (amount && amount > 125) {
     throw  new Error(ERROR_MESSAGES.MAX_AMOUT);
```
tifrel: autofmt broken?

packages/sdk/src/mint/mint.ts
```diff
@@ -36,80 +35,122 @@ export const mint = (
...
   }
 
-  if (royaltyPercentage && royaltyPercentage < 0 || royaltyPercentage > 0.5) {
-    throw new Error(ERROR_MESSAGES.INVALID_ROYALTY_PERCENTAGE);
+  //if specifying tokenIdsToMint these must be populated
+  if (tokenIdsToMint && tokenIdsToMint.length == 0) {
+    throw  new Error(ERROR_MESSAGES.EMPTY_TOKEN_IDS);
```
tifrel: autofmt broken?

packages/sdk/src/mint/mint.ts
```diff
@@ -36,80 +35,122 @@ export const mint = (
...
   }
 
+  if (tokenIdsToMint && amount && tokenIdsToMint?.length !== amount) {
+    throw new Error (ERROR_MESSAGES.MUTUAL_EXCLUSIVE_AMOUNT);
+  }
+  const adjustedTokenCount = tokenIdsToMint?.length ? tokenIdsToMint?.length : amount;
```
tifrel: Isn't this `tokenIdsToMint?.length || amount`?

packages/sdk/src/mint/mint.ts
```diff
@@ -36,80 +35,122 @@ export const mint = (
...
+  //if no amount or tokenids defined then default to 1 amount
+  const adjustedAmount = !amount && !tokenIdsToMint ? 1 : adjustedTokenCount;
```
tifrel: And with the simplification above, this could be `tokenIdsToMint?.length || amount || 1`, right? (Should probably be verified by tests)

packages/sdk/src/mint/mint.ts
```diff
@@ -36,80 +35,122 @@ export const mint = (
...
     args: {
       owner_id: ownerId,
       metadata: metadata,
-      num_to_mint: amount || 1,
+      num_to_mint: adjustedAmount,
       // 10_000 = 100% (see above note)
-      royalty_args: !splits ? null : { split_between: splits, percentage: royaltyPercentage * 10_000 },
-      split_owners: splits || null,
-      token_ids_to_mint: tokenIdsToMint,
+      royalty_args: royaltyTotal < 0 || !royaltyTotal ? null : { split_between: roundedRoyalties, percentage: Math.floor(royaltyTotal * 10000) },
```
tifrel: How should royaltyTotal ever be less than null? I hope we are checking that people don't try negative percentages here.

packages/sdk/src/mint/mint.ts
```diff
@@ -36,80 +35,122 @@ export const mint = (
...
 };
 
-function adjustSplitsForContract(splits: Record<string, number> ): void {
+function getRoyaltyTotal(royalties: Record<string, number> ): number {
+  let royaltyTotal = 0;
+  Object.values(royalties).forEach(value => {
+    royaltyTotal += value;
+  });
+
+  if (royaltyTotal <= 0 || royaltyTotal > 0.5) {
```
tifrel, line 102: So here we have the validation, such that the case of a negative royaltyTotal cannot happen.

packages/sdk/src/mint/mint.ts
```diff
@@ -36,80 +35,122 @@ export const mint = (
...
+    if (royalties[key] <= 0) {
+      throw new Error (ERROR_MESSAGES.NEGATIVE_ROYALTIES);
+    }
+    const adjustedAmount = royalties[key]/ royaltyTotal * 10000;
+    result[key] = adjustedAmount;
+    counter += adjustedAmount;
   });
-  if (counter != 1) {
+  if (counter != 10000) {
     throw new Error (ERROR_MESSAGES.SPLITS_PERCENTAGE);
```
tifrel: Shouldn't it be a royalties percentage message? This message will not help mbjs users

packages/sdk/src/mint/mint.ts
```diff
@@ -36,80 +35,122 @@ export const mint = (
...
   }
+ 
+  return result;
+}
+
+function roundRoyalties(royalties: Record<string, number>): Record<string, number> {
```
tifrel, line 126: Is this well-tested with automated tests?

packages/sdk/src/mint/mint.ts
```diff
@@ -36,80 +35,122 @@ export const mint = (
...
+
+  if (roundedCounter != 10000) {
+    result[firstKey] += 10000 - roundedCounter;
+  }
+  return result;
 }
 
 function mintingDeposit({
   nTokens,
-  nSplits,
```
tifrel, line 96: We should keep this in the minting deposit, both for completeness and reference, though we will still need to fix this method.

Also we'll have to take the required deposit into account when doing the `setSplitsAndList` method

packages/sdk/src/mint/mint.ts
```diff
@@ -36,80 +35,122 @@ export const mint = (
...
 }: {
   nTokens: number;
-  nSplits: number;
   nRoyalties: number;
   metadata: TokenMetadata;
 }): string {
   const commonDeposit = new BN(DEPOSIT_CONSTANTS.STORE_COMMON);
   const royaltiesDeposit = commonDeposit.mul(new BN(nRoyalties));
-  const splitsDeposit = commonDeposit.mul(new BN(nSplits));
+  const splitsDeposit = commonDeposit.mul(new BN(nRoyalties));
```
tifrel: Why would we just double the royalties deposit here?

packages/sdk/src/mint/mint.ts
```diff
@@ -36,80 +35,122 @@ export const mint = (
...
   const mintingFee = new BN(MINTING_FEE);
 
   // JSON serialization should give us an estimate that's always higher than
   // borsh serialization
-  const metadataDeposit = new BN(YOCTO_PER_BYTE).mul(new BN(JSON.stringify(metadata).length));
+  const metadataDeposit = new BN(YOCTO_PER_BYTE).mul(new BN(JSON.stringify(metadata)?.length));
```
tifrel: Good point. But even empty/undefined metadata has a base cost.

packages/sdk/src/mint/mint.ts
SurgeCode: Good point, changing this

packages/sdk/src/mint/mint.test.ts
SurgeCode: Added it in somewhere else

packages/sdk/src/mint/mint.ts
SurgeCode: Need to verify floating points before adding to complete rounded value

packages/sdk/src/mint/mint.ts
SurgeCode: Done

packages/sdk/src/mint/mint.ts
SurgeCode: Throws error now

packages/sdk/src/errorMessages.ts
SurgeCode: Think we found that it would not have the problems we anticipated in the end so doing 0 - 0.5 is ok

packages/sdk/src/mint/mint.test.ts
SurgeCode: Added

packages/sdk/src/mint/mint.ts
SurgeCode: Done

packages/sdk/src/mint/mint.ts
SurgeCode: Yes, tried to cover some edgecases

packages/sdk/src/mint/mint.ts
```diff
@@ -36,81 +35,125 @@ export const mint = (
...
   }
 
-  if (royaltyPercentage && royaltyPercentage < 0 || royaltyPercentage > 0.5) {
-    throw new Error(ERROR_MESSAGES.INVALID_ROYALTY_PERCENTAGE);
+  if (tokenIdsToMint && amount && tokenIdsToMint?.length !== amount) {
+    throw new Error(ERROR_MESSAGES.MUTUAL_EXCLUSIVE_AMOUNT);
   }
 
+  //if no amount or tokenids defined then default to 1 amount
+  const adjustedAmount = tokenIdsToMint?.length || amount || 1;
```
tifrel, line 69: So if I specify 1000 token IDs, mbjs will let me try to mint 1000 tokens, right?

packages/sdk/src/mint/mint.ts
```diff
@@ -36,81 +35,125 @@ export const mint = (
...
     args: {
       owner_id: ownerId,
       metadata: metadata,
-      num_to_mint: amount || 1,
+      num_to_mint: adjustedAmount,
       // 10_000 = 100% (see above note)
-      royalty_args: !splits ? null : { split_between: splits, percentage: royaltyPercentage * 10_000 },
-      split_owners: splits || null,
-      token_ids_to_mint: tokenIdsToMint,
+      royalty_args: !royaltyTotal ? null : { split_between: roundedRoyalties, percentage: Math.floor(royaltyTotal * 10000) },
```
tifrel: Any reason for flooring div over rounding?

packages/sdk/src/mint/mint.ts
```diff
@@ -36,81 +35,125 @@ export const mint = (
...
+  const result: Splits = {};
+  const firstKey = Object.keys(royalties)[0];
+  Object.keys(royalties).forEach((key) => {
+    const roundedVal = Math.round(royalties[key]);
+    result[key] = roundedVal;
+    roundedCounter += roundedVal;
+  });
+
+  if (roundedCounter != 10000) {
+    result[firstKey] += 10000 - roundedCounter;
```
tifrel, line 137: This is actually the smart way of handling this ;)

packages/sdk/src/errorMessages.ts
```diff
@@ -3,16 +3,23 @@ const SUPPORT = '- further help available on our telegram channel: https://t.me/
...
   TOKEN_NOT_ARRAY: `tokenIds must be an array ${SUPPORT}`,
   TOKEN_LENGTH: `tokenIds must have more than one element ${SUPPORT}`,
-  TRANSFER:  `You must transfer at least one token ${SUPPORT}`,
+  TRANSFER: `You must transfer at least one token ${SUPPORT}`,
   NO_MINTERS: `There are no minters being provided to the batchChangeMinters method to be changed, try adding account Ids to at least one of the minter fields ${SUPPORT}`,
   NO_REFERENCE: `You must provide a reference when minting or explicitly opt out ${SUPPORT}`,
   NO_MEDIA: `You must provide media when minting or explicitly opt out ${SUPPORT}`,
+  ROYALTY_PERCENTAGE_WITHOUT_RECIPIENTS: `Cannot specifiy royalty percentage without splits or royalties ${SUPPORT}`,
+  ROYALTY_RECIPIENTS_WITHOUT_PERCENTAGE: `Cannot specify royalties without royalty percentage ${SUPPORT}`,
+  TOKEN_IDS_TO_MINT_LEN: `Token ids to mint must not be included or have a length greater than 0 ${SUPPORT}`,
```
tifrel: Is this error message dead code?

packages/sdk/src/mint/mint.ts
```diff
@@ -36,81 +35,125 @@ export const mint = (
...
+    royaltyTotal = getRoyaltyTotal(royalties);
+    adjustedRoyalties = adjustRoyaltiesForContract(royalties, royaltyTotal);
+    roundedRoyalties = roundRoyalties(adjustedRoyalties);
   }
 
-  if (splits && Object.keys(splits).length > 50) {
-    throw new Error(ERROR_MESSAGES.MAX_SPLITS);
+  //if royalties exist they need to be populated
+  if (royalties && Object.keys(royalties).length < 1) {
+    throw new Error(ERROR_MESSAGES.MIN_ROYALTIES);
```
tifrel, line 51: Also mustn't be more than 50

packages/sdk/src/mint/mint.ts
SurgeCode: We can do round

SurgeCode (approved):

sainthiago (approved):

packages/sdk/src/mint/mint.ts
```diff
@@ -36,86 +35,130 @@ export const mint = (
...
+  const result: Splits = {};
+  const firstKey = Object.keys(royalties)[0];
+  Object.keys(royalties).forEach((key) => {
+    const roundedVal = Math.round(royalties[key]);
+    result[key] = roundedVal;
+    roundedCounter += roundedVal;
+  });
+
+  if (roundedCounter != 10000) {
+    result[firstKey] += 10000 - roundedCounter;
```
cif, line 137: I was looking for this line :)

cif (approved):

PR #313, opened 2023-02-22
feat(sdk): :sparkles: Prevent splits from being passed

cif:
Precursor to work @SurgeCode is doing. We have a use case we only want royalties now.


sainthiago (approved):

cif:
Closing this in favor of HEAD / current development, will bring only consumer (demo) up to speed

PR #314, opened 2023-02-23
setup all wallets

rubenmarcus:


cif:
Does this create need to test more of the wallets or this is a revert?  

Might be nice to offer mbjs users a way to pass configs or even require the modules on the consuming side, maybe consider dynamic imports if target frameworks support it

sainthiago (approved):

cif:
This is starting to get stale and we are going to be more methodical about this rather than just adding all wallets across the board

PR #330, opened 2023-03-16
FT integration: USDC

tifrel:


cif (approved):
From client perspective this looks pretty straight forward. Assume users will be able to create two listings in each token if they would like to or will market contract state prevent this from being possible?  e.g replace a NEAR listing with nUSDC?

sainthiago (approved):

PR #335, opened 2023-04-03
fix(data): :wrench: Fix codegen config

cif:


packages/data/src/graphql/codegen/index.ts
```diff
@@ -1,2 +1,2 @@
-export * from './gql';
-export * from './fragment-masking';
+export * from "./gql"
+export * from "./fragment-masking"
```
nategeier, line 2: are we using double quotes now?

rubenmarcus (approved):

nategeier (changes requested):

sainthiago (approved):

nategeier (approved):

PR #338, opened 2023-04-12
test(testing): :alembic: Seed phrase and zero balance account

cif:
This is an integration test which could be encapsulated into a mbjs method which produced a zero balance (sub account), which when paired with meta transactions in [near-api-js upcoming releases](https://github.com/near/near-api-js/pull/1097) could allow users to connect to services without having to purchase any NEAR token.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5914,6 +5914,158 @@
         "node": ">=4"
       }
     },
+    "node_modules/@near-js/accounts": {
+      "version": "0.0.2",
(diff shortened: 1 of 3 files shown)


cif:
There isn't much of a need to merge this one really, think we'll rely on wallets to create the accounts for meta transactions

sainthiago (approved):

cif:
This was only for reference, no need to merge this

PR #339, opened 2023-04-17
feat(auth): :sparkles: Add meteor validations to wallet auth

cif:
- Add `isMeteor` method to auth module (can be consumed directly w/o context)
- Add `getSelectorState` method to auth module (can be consumed directly w/o context)
- Add `verifyMessage` method to auth module

### DIFF
--- a/packages/app/pages/_app.tsx
+++ b/packages/app/pages/_app.tsx
@@ -3,6 +3,10 @@ import type { AppProps } from 'next/app';
 import { WalletContextProvider } from '@mintbase-js/react';
 import '@near-wallet-selector/modal-ui/styles.css';
 
+globalThis.mbjs.network = 'mainnet';
+globalThis.mbjs.mbContract = 'mintbase1.near';
+globalThis.mbjs.contractAddress = 'mintbase1.near';
+
 function MyApp({ Component, pageProps }: AppProps): JSX.Element {
   return (
     <WalletContextProvider>
--- a/packages/app/pages/index.tsx
+++ b/packages/app/pages/index.tsx
@@ -1,11 +1,11 @@
 import { useWallet } from '@mintbase-js/react';
-import { getSelectorState, isMeteorWallet, verifyMessage } from '@mintbase-js/auth';
+// import { getSelectorState, isMeteorWallet, verifyMessage } from '@mintbase-js/auth';
 import type { NextPage } from 'next';
 import Head from 'next/head';
 import { TokenExample } from '../components/TokenExample';
 import { TransferTest } from '../components/TransferTest';
 import styles from '../styles/Home.module.css';
-import { useState } from 'react';
+
 
 const Home: NextPage = () => {
   const {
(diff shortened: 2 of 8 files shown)


packages/app/pages/_app.tsx
```diff
@@ -3,6 +3,10 @@ import type { AppProps } from 'next/app';
 import { WalletContextProvider } from '@mintbase-js/react';
 import '@near-wallet-selector/modal-ui/styles.css';
 
+globalThis.mbjs.network = 'mainnet';
```
cif, line 6: I think we may still have something missing on config side, but can worry about that later

packages/app/pages/index.tsx
```diff
@@ -18,11 +19,13 @@ const Home: NextPage = () => {
   } = useWallet();
 
   const signMessageTest = async (): Promise<void> => {
-    await signMessage({
+    console.log('is meteor?', isMeteorWallet(), getSelectorState());
```
cif: vou limpiar

packages/auth/src/wallet.test.ts
```diff
@@ -186,4 +187,27 @@ describe('wallet', () => {
...
+    'accountId': 'meteor_with_no_monies.near',
+    'message': 'hey',
+    'blockId': 'FXbpnvguB1ETCwaFk4beHgsekHBf1JkntPBM8MkWYw6g',
+    'publicKey': 'PXwxiJEOmsHAasQvHywXV1CDunJvj9K2DHGo+I8icA0=',
+    'keyType': 0,
+    'signature': '8YKOezp2v/6iiKtGIRYy/U1DhH43ZJWSBqRtxT8HlYqeMAES05kpL6H2XffFF/YaQTMrQDcatTIt2T269qPtDg==',
+  };
+
+  test('verify valid message', () => {
+    const result = verifyMessage(VALID_MESSAGE_PAYLOAD);
```
cif, line 202: worth doing more testing in the future on what aspects of this message matter, right now the whole thing is stringified, was taken from @microchipgnu implementation in api layer on omni site

microchipgnu (approved):

sainthiago (approved):

packages/auth/src/wallet.test.ts
cif: Unfortunately the isomorphic encoder seems to be producing a different byte array in the browser, but for now we should only need this server side anyway

packages/auth/src/wallet.ts
```diff
@@ -176,29 +197,47 @@ export const getVerifiedOwner =
...
-// https://github.com/near/wallet-selector/issues/434
-// export const verifyMessage = async (signature: string): Promise<boolean> => {
-
-//   // const owner = await getVerifiedOwner(signature);
-
-//   // const publicKeyString = `ed25519:${BinaryToBase58(Buffer.from(owner.publicKey, 'base64'))}`;
+export const verifyMessage = (payload: SigningPayload): boolean => {
+  // for now, just warn in browser
+  // isomporphic result produces different encoding byte lengths
+  if (typeof window !== 'undefined') {
```
cif, line 218: Adding this warning to deal with issue in test comment

rubenmarcus (approved):

PR #346, opened 2023-04-28
refactor(auth): :lock: Refactor auth module for better security

cif:
Wraps calls to connect service in convenient locations for use in other service and/or proxy.

### DIFF
--- a/packages/app/pages/index.tsx
+++ b/packages/app/pages/index.tsx
@@ -47,6 +47,7 @@ const Home: NextPage = () => {
 
   const testFirebaseLogin = async (): Promise<void> => {
     const auth = getAuth();
+    console.log('got session?', session);
     const login = await signInWithCustomToken(auth, session.token);
     console.log('has firebase login!', login);
   };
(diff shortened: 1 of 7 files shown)


packages/auth/src/session.ts
```diff
@@ -0,0 +1,34 @@
...
+      'mb-api-key': 'anon',
+      'content-type':'application/json',
+    },
+    body: JSON.stringify(req.body),
+  });
+  const { token } = await session.json();
+  return token;
+};
+
+export const validateSessionToken = async (token: string, useEndpoint?: string): Promise<MintbaseSession> => {
```
cif, line 22: This method in particular can be used by any endpoint, anywhere in the services layer including vercel endpoints.

packages/app/pages/index.tsx
```diff
@@ -47,6 +47,7 @@ const Home: NextPage = () => {
 
   const testFirebaseLogin = async (): Promise<void> => {
     const auth = getAuth();
+    console.log('got session?', session);
```
cif, line 50: This is probably fine for internal test client. Demoed to @SurgeCode and @microchipgnu

rubenmarcus (approved):

PR #369, opened 2023-07-11
:package: Remove unused isomorphic fetch lib

sainthiago:


tifrel (approved):

tifrel (changes requested):
This will break things, see @rubenmarcus comments in the dev chat

packages/data/package.json
```diff
@@ -21,8 +21,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "graphql-request": "^5.0.0",
-    "isomorphic-fetch": "^3.0.0"
```
rubenmarcus, line 25: it is needed to server side

rubenmarcus (changes requested):

PR #376, opened 2023-08-03
feat (wallet): add package

microchipgnu:


packages/react/src/WalletContext.tsx
```diff
@@ -115,6 +122,18 @@ export const WalletContextProvider: React.FC<{ children: React.ReactNode; networ
     };
   }, [setup]);
 
+
+  useEffect(() => {
+    // Check if mintbasewallet:activeAccountId exists in localStorage
+    const activeAccountId = localStorage.getItem('mintbasewallet:activeAccountId');
+
```
microchipgnu: Not good to have this hardcoded. This solution will only work for our wallet implementation.

PR #394, opened 2023-09-11
fix beta branch build

sainthiago:


GAllen98 (approved):

tifrel (commented):
Why is this such a big downgrade? According to npmjs our last released version is 0.8.5

PR #403, opened 2023-10-06
Add mb wallet multiple txns

rubenmarcus:


packages/auth/src/wallet.ts
```diff
@@ -35,11 +35,18 @@ export let walletSelectorComponents: WalletSelectorComponents  = {
...
   const selector = await setupWalletSelector({
     network: network,
     debug: mbjs.keys.debugMode,
     modules: [
-      ...options?.additionalWallets || [],
+      setupMBWallet({
+        networkId: network,
+        relayerUrl: '/api/relay',
+        signInContractId: MINTBASE_CONTRACTS[network],
+        walletUrl: network ==='testnet' ? 'https://testnet.wallet.mintbase.xyz' : 'https://testnet.wallet.mintbase.xyz',
```
bh2smith, line 48: It looks to me like the wallet url is testnet regardless of the network value? Shouldn't this be more like:

```suggestion
        walletUrl: network === 'testnet' ? 'https://testnet.wallet.mintbase.xyz' : 'SomethingDifferent',
```

packages/react/src/WalletContext.tsx
```diff
@@ -68,6 +68,8 @@ export const WalletContextProvider: React.FC<{ children: React.ReactNode; networ
   const selectedContract = contractAddress || mbjs.keys.contractAddress;
 
   const setup = useCallback(async () => {
+
+    console.log('setup')
```
bh2smith, line 72: nit: Seems like this wasn't intended to keep:

```suggestion
```

packages/sdk/src/execute/execute.ts
```diff
@@ -21,7 +21,11 @@ export const execute = async (
   validateSigningOptions({ wallet, account });
 
   const outcomes = await genericBatchExecute(flattenArgs(calls), wallet, account, callbackUrl, callbackArgs);
+  console.log(calls, 'calls')
 
-  return checkCallbackUrl(callbackUrl, callbackArgs, wallet, outcomes);
+  console.log(outcomes, 'outcomes')
+
+
+  // return checkCallbackUrl(callbackUrl, callbackArgs, wallet, outcomes);
```
bh2smith, line 29: This looks forgotten...
```suggestion
```

bh2smith (commented):
This looks like it is not ready for review. There are a lot of commented code blocks console logs that appear to be for local debugging purposes... I started off by making suggestions, but later realized there were many more.

PR #410, opened 2023-10-12
RPC: Get access keys for account

tifrel:


sainthiago (approved):

packages/rpc/src/methods/keys.ts
```diff
@@ -0,0 +1,36 @@
...
+    allowance: string;
+    receiver_id: string;
+    method_names: string[];
+  };
+}
+
+export const getAccessKeys = async (accountId: string): Promise<AccessKey[]> => {
+  const res = await requestFromNearRpc({
+    jsonrpc: '2.0',
+    id: 'dontcare',
```
bh2smith, line 19: I suppose this is a meaningless field and we can send whatever we want.

packages/rpc/src/methods/keys.test.ts
```diff
@@ -0,0 +1,22 @@
+import fetch from 'isomorphic-unfetch';
+import { getAccessKeys } from './keys';
+
+jest.mock('isomorphic-unfetch');
+
+describe('keys', () => {
```
bh2smith, line 6: Could we also add tests that the errors are thrown when the response fields don't exist? That way we would know if someone accidentally deletes one of those "throws".

bh2smith (approved):
LGTM! Although I think those two extra tests could help out in the future.

packages/rpc/src/methods/keys.ts
tifrel: I just checked the source code and it seems they simply copy it over to the response, so I guess you could use it to order things if you have an async system running, other folks might come up with more fancy use cases ;)

packages/rpc/src/methods/keys.test.ts
tifrel: Yup, done

PR #409, opened 2023-10-12
@mintbase-js/wallet

rubenmarcus:

PR #413, opened 2023-10-18
Revert "@mintbase-js/wallet"

sainthiago:
Reverts Mintbase/mintbase-js#409


tifrel (approved):

bh2smith:
Why tho?

sainthiago:
> Why tho?

@bh2smith as @rubenmarcus explained to me, this was created to only have mintbase wallet in the wallet list. so this is only suitable for minsta. we will improve this in the next days before merge (allow user to choose which wallets he wants in the consumer app).

PR #415, opened 2023-10-19
Removed at must be `NULL`

tifrel:


sainthiago (approved):

tmp.sh
```diff
@@ -0,0 +1,23 @@
+NODE_URL='https://rpc.ankr.com/eth'
```
bh2smith, line 1: I think change is not intended to be here.

bh2smith (commented):
I believe you have accidentally committed a shell script for converting (block_number, transaction_index) to transaction hash here.

PR #434, opened 2023-11-16
Add async keyword for signAndSendTransaction

kujtimprenkuSQA:
# Description

This small change is needed because `signAndSendTransaction` is an `async` function for all wallet types.

One of the reasons this is needed is the fact that the code is running `synchronously` for this function and the `.then` or `.catch` in dApps gets ignored when `signAndSendTransaction` is called.

---
This change is required because if `signAndSendTransaction` throws or returns anything it just shows in console but it does not get returned to the dApp.

mintbase-signAndSendTransaction-error

---

After this change we can catch and show the error to users:

image

### DIFF
--- a/packages/wallet/src/mintbase-wallet.ts
+++ b/packages/wallet/src/mintbase-wallet.ts
@@ -172,7 +172,7 @@ export const MintbaseWallet: WalletBehaviourFactory<
     return;
   };
 
-  const signAndSendTransaction = ({
+  const signAndSendTransaction =  async ({
     receiverId,
     actions,
     signerId,


kujtimprenkuSQA:
Hi, @rubenmarcus we need to make this small fix, apologies for not noticing it earlier today.

microchipgnu (approved):

microchipgnu:
LGTM!

sainthiago (approved):

rubenmarcus:
> Hi, @rubenmarcus we need to make this small fix, apologies for not noticing it earlier today.

approved!

PR #447, opened 2023-12-29
Remove addMinter and removeMinter methods in sdk

sainthiago:


packages/sdk/src/addMinter/addMinter.ts
```diff
@@ -1,29 +0,0 @@
...
-  if (contractAddress == null) {
-    throw new Error(ERROR_MESSAGES.CONTRACT_ADDRESS);
-  }
-  
-  return {
-    contractAddress: contractAddress || mbjs.keys.contractAddress,
-    args: {
-      account_id: minterId,
-    },
-    methodName: TOKEN_METHOD_NAMES.ADD_MINTER,
```
tifrel, line 25: Shouldn't these method names be deleted as well then?

tifrel (approved):

packages/sdk/src/addMinter/addMinter.ts
sainthiago: you are right! 
updated.

PR #454, opened 2024-01-19
Add contractMinters method in data package

sainthiago:


packages/data/src/api/contractMinters/contractMinters.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ParsedDataReturn } from '../../types';
+import { parseData } from '../../utils';
+import { contractMintersQuery } from './contractMinters.query';
+import { ContractMintersResults } from './contractMinters.types';
+
+
+export const contractMinters = async (
+  contractAddress: string | string[],
+  network?: Network,
+): Promise<ParsedDataReturn<ContractMintersResults>> => {
```
tifrel: It would be more straight-forward for users if they just got a `string[]` as return.

packages/data/src/api/contractMinters/contractMinters.ts
```diff
@@ -0,0 +1,24 @@
+import { Network } from '@mintbase-js/sdk';
+import { fetchGraphQl } from '../../graphql/fetch';
+import { ParsedDataReturn } from '../../types';
+import { parseData } from '../../utils';
+import { contractMintersQuery } from './contractMinters.query';
+import { ContractMintersResults } from './contractMinters.types';
+
+
+export const contractMinters = async (
+  contractAddress: string | string[],
```
tifrel: Does `string[]` work? I would assume that it crashes the query. Also what is returned then, a union of minters? Are duplicates removed? This would be a lot less confusing if we dropped to possibility to query multiple smart contracts at once.

tifrel (changes requested):

tifrel (approved):

PR #481, opened 2024-03-07
feat(data): Add reference field to store NFTs

tifrel:


bh2smith (approved):
Seems legit. Surprised this wasn't here before... or what was wrong with the "image" filed that seems to hold the same info?

sainthiago (approved):

PR #483, opened 2024-03-08
Implement dynamic NFT metadata query

tifrel:


packages/data/src/api/dynamicNftMetadata/README.md
```diff
@@ -0,0 +1,33 @@
...
+For the most reliable data, reference our [existing graphql docs](https://docs.mintbase.io/dev/read-data/mintbase-graph).
+{% endhint %}
+
+Returns NFTss by `contractAddress` and `internMetadataId` (as used on the smart contract). The `isActive` toggle filters for active (currently used) or inactive (unused) metadata.
+
+### dynamicNftMetadata( {contractAddress: string, internalMetadataId: number, isActive: boolean, network?: "testnet" | "mainnet"})
+
+Example usage:
+
+{% code title="queryDynamicNftMetadata.ts" overflow="wrap" lineNumbers="true" %}
```
bh2smith, line 16: Today I learned this little trick!

packages/data/src/api/dynamicNftMetadata/dynamicNftMetadata.mock.ts
```diff
@@ -0,0 +1,55 @@
...
+    {
+      metadataId: 'dynamic.mintbase2.near:deadbeef0',
+      contractAddress: 'dynamic.mintbase2.near',
+      internalMetadataId: 0,
+      reference: 'https://arweave.net/reference0',
+      referenceBlob: {
+        media: 'https://arweave.net/media0',
+        history: [],
+      },
+      isActive: true,
```
bh2smith: This mock value is called inactive, but has active = true... was that intended?
```suggestion
      isActive: false,
```

packages/data/src/api/dynamicNftMetadata/dynamicNftMetadata.test.ts
```diff
@@ -0,0 +1,62 @@
...
+  it('should return inactive metadata', async () => {
+    (GraphQLClient as jest.Mock).mockImplementationOnce(() => ({
+      request: (): Promise<DynamicNftMetadataResult> => Promise.resolve(dynamicNftMetadataMockInactive),
+    }));
+
+    const result = await dynamicNftMetadata({ ...props, isActive: false });
+
+    expect(result?.data).toBe(
+      dynamicNftMetadataMockInactive,
+    );
```
bh2smith, line 48: If this is passing and the inactiveMock has only active records, I am not sure I understand...

sainthiago (approved):

packages/data/src/api/dynamicNftMetadata/dynamicNftMetadata.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  it('should handle errors', async () => {
+    const errMessage = 'exploded';
+    (GraphQLClient as jest.Mock).mockImplementationOnce(() => ({
+      request: (): Promise<DynamicNftMetadataResult> => Promise.reject(new Error(errMessage)),
+    }));
+
+    const call = await dynamicNftMetadata({...props, isActive: true});
+
+    expect(call).toStrictEqual({ error: errMessage });
```
bh2smith, line 59: Here, I think we should be mocking the internal `fetchGraphQl` to actually test that `dynamicNftMetadata` is handling errors the way we expect and the assertion should look like:

```suggestion
    expect(call).toStrictEqual({ error: `Error fetching store nfts, ${errMessage}` });
```

Because that's what the function does with errors. 

Otherwise, I don't believe this test serves any non-trivial purpose.

bh2smith (commented):
Everything looks overall good except for two few suspicious things inside the tests:

1. mocked inactive value has `active = true` which seems unintended (and also weird if tests are passing on it). 
2. The test for error capturing seems overly mocked to the point that it is not actually testing anything.

rubenmarcus (approved):

packages/data/src/api/dynamicNftMetadata/dynamicNftMetadata.test.ts
tifrel: I do agree that the tests are a bit over-mocked, this is essentially just testing that the function forwards the GraphQL result as the data.

packages/data/src/api/dynamicNftMetadata/dynamicNftMetadata.test.ts
tifrel: Hm, the method `GraphQlClient.request` is being used inside of `fetchGraphQl`, so this is indeed a bit spurious. Just dug into `parseData`, the `Error fetching store NFTs` is simply logged, but not returned. it returns the unchanged error string or object. To change that would require rewriting the tests for the whole module.

bh2smith (approved):

PR #499, opened 2024-03-27
update mbjs-react

rubenmarcus:


packages/react/README.md
```diff
@@ -60,7 +60,7 @@ the default way of interacting with Mintbase Wallet is using the MintbaseWalletC
 
 ## properties:
 
-**contractAddress** : `your near contract address / your mintbase store address`
+**contractAddress** (not mandatory): `If you set this it will login the user using Limited Access Keys,set with your near contract address / your mintbase store address`
```
sainthiago:
```suggestion
**contractAddress** (optional): `If you set this it will connect the user using Limited Access Keys, set with your near contract address / your mintbase store address`
```

GAllen98: I would change the order of the scentences like so: 
`Your near contract address / your mintbase store address. If you set this, it will connect the user using Limited Access Keys.`

sainthiago (approved):

PR #506, opened 2024-04-29
Fix decimals for FT minting

tifrel:


packages/sdk/src/mintOnMetadata/mintOnMetadata.ts
```diff
@@ -37,6 +39,10 @@ export const mintOnMetadata = (
     throw new Error(ERROR_MESSAGES.METADATA_ID_NOT_INT);
   }
 
+  if ((ftAddress && !ftDecimals) || (ftDecimals && !ftAddress)) {
```
sainthiago, line 42: should we add ftAddress and ftDecimals mandatory in the type then?

tifrel: The default case is that neither are given, and the payment is then done with $NEAR.

sainthiago (approved):

PR #515, opened 2024-05-28
Update near-api-js to v3

bh2smith:
This PR

1. 52c4ce736bff0581b1e415afc21ae3589bf4791d: removes a few unused dev-dependencies and imports.
2. 5c6f3c2405b8f80a0b01f3d05ae8909fc9478c32: deletes the duplicated nearPrice file (since it should have just been imported). 
3. 04584f43f53aacc8dee2ff1177d5524970f6d7fd: Update near-api-js everywhere
4. 26ad431b6dd829c4b2695b170e4a4a5766ef79fc: Updates lerna because of some build error

The last few commits are just fixing CI errors (note that I am receiving inconsistent result logs for linter locally vs in CI)

## Test Plan

Existing CI (i.e. project builds, is linted and tests pass).


sainthiago (approved):

bh2smith:
This is a bit old and stale. Do we actually want to make this change? It would help a lot in the long run, but not sure how it will affect the system. Also, we would want to proceed with v4 soon after.

bh2smith:
Oh -- its just mintbase-js. I guess this is fine because its a beta branch
Are we using the beta branch anywhere?

sainthiago:
> Oh -- its just mintbase-js. I guess this is fine because its a beta branch Are we using the beta branch anywhere?

yes our beta branch is used everywhere.

rubenmarcus:
This is gonna be force-reverted

PR #518, opened 2024-07-03
Upgrade NearApiJS to v4

bh2smith:
There was only one snag in the way we convert wallet-selector.Action into nearApiJs.Action where we were previously relying on the wallet-selector to do the conversion (it is outdated) but nearApiJs has a supportive constructor.


.github/workflows/branch-publish.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         run: |
           echo NEAR_NETWORK=testnet >> .env
           echo NFT_CONTRACT_ID=default_contract >> .env
-          npm ci
+          npm install --frozen-lockfile
```
bh2smith:
```suggestion
          npm ci
```

packages/auth/src/index.ts
```diff
@@ -10,4 +10,4 @@ export { Account, providers, connect as connectToNear, KeyPair } from 'near-api-
 export { InMemoryKeyStore, KeyStore } from 'near-api-js/lib/key_stores';
 
 // adding some other types from within libraries as a convenience
-export { FinalExecutionOutcome } from '@near-wallet-selector/core';
```
tifrel, line 13: Below, njs.FEO has been exchanged for nws.FEO. Why is it the other way around here?

packages/sdk/src/execute/execute.utils.ts
```diff
@@ -121,8 +120,8 @@ const batchExecuteWithNearAccount = async (
           contractId: call.contractAddress,
           methodName: call.methodName,
           args: call.args,
-          gas: new BN(call.gas),
-          attachedDeposit: new BN(call.deposit),
+          gas: BigInt(call.gas.toString()),
```
tifrel: Shouldn't `call.gas` be a string or Bigint already?

packages/sdk/src/types.ts
```diff
@@ -520,7 +519,7 @@ export type ExecuteArgsResponse = TransferArgsResponse
   | BatchChangeMintersArgsResponse
   | Record<string, unknown>;
 
-export type FinalExecutionOutcome = FEO;
+export type FinalExecutionOutcome = providers.FinalExecutionOutcome;
```
tifrel, line 521: Here moving from nws to naj again?

packages/sdk/src/execute/execute.utils.ts
tifrel: Also, shouldn't this be totally fine with taking a string instead of casting to BigInt again? Pretty sure it's sent to RPC as a string.

tifrel (changes requested):

packages/auth/src/index.ts
bh2smith: I would argue that is it not the other way around here. It is the same way around. That is, in both cases we export from `near-api-js`.

In any case, it seems that we have to maintain their version of this type until they update the near-api on their end. A least in a couple of places. Perhaps in the places where it is still being used, I can give it a different name, but I don't understand why you think here is "the other way around" - maybe take another look at that.

packages/sdk/src/execute/execute.utils.ts
bh2smith: Perhaps the issue here is that we are still using the "Old Type". Will check.

> Also, shouldn't this be totally fine with taking a string instead of casting to BigInt again? 

I was just following the compiler errors.

> Shouldn't call.gas be a string or Bigint already?

call.gas here is of type `string | BN` and it is expected to be `bigint`. However, we do own this type `ContractCall` we could change it at the source.

Screenshot 2024-07-03 at 21 46 14

packages/sdk/src/types.ts
bh2smith: I was just trying to only use the one type everywhere. Realistically, I don't know why we are re-exporting this type. Is it intended that users of this package wouldn't also depend on near-api or wallet-selector?

bh2smith (commented):
Thanks for your eyes @tifrel - I have now explicitly labeled places where we use the NWS execution outcome. 

These two types are coming from @near-js (which is a dependency of @near-api-js) which is older on the wallet-selector because its stuck back at `near-api-js` v2.

This is the OLD TYPE:
Screenshot 2024-07-03 at 22 08 22

This is the NEW Type:

Screenshot 2024-07-03 at 22 09 34

I would love to just "cast" the data as the new type in all places that are user facing and expect it the missing field to be there (since RPCs have likely been updated to include them anyway). The we wouldn't be leaking the old type anywhere in a user facing interface and can essentially phase out the old type.

What do you think. I could probably even make a test that casting Old as New results in having that missing field `final_execution_status` (unless the wallet selector explicitly deletes it from the response data, it should still be there).

bh2smith:
Closed in favour of #532

PR #519, opened 2024-07-05
Use `mintbasev2.near` instead of `mintbase2.near`

tifrel:


sainthiago (approved):

packages/data/src/api/dynamicNftMetadata/README.md
```diff
@@ -18,7 +18,7 @@ Example usage:
 import { dynamicNftMetadata } from  '@mintbase-js/data'
 
 const props = {
-  contractAddress: 'dynamic.mintbase2.near',
+  contractAddress: 'dynamic.mintbasev2.near',
```
bh2smith, line 21: This appears so many times it could easily be a global constant. Then future changes wouldn't touch so much code.

packages/sdk/src/mint/mint.test.ts
```diff
@@ -6,7 +6,7 @@ import { mbjs } from '../config/config';
 
 describe('mint method tests', () => {
   const contractAddress = `test.${mbjs.keys.mbContract}`;
-  // const contractAddressV2 = 'test.mintbase2.near';
+  // const contractAddressV2 = 'test.mintbasev2.near';
```
bh2smith, line 9:
```suggestion
```

bh2smith (approved):
Please make this value a global constant.

sainthiago (approved):

packages/data/src/api/dynamicNftMetadata/README.md
tifrel: It already is a global constant, but used only sporadically. The changes also resulted from the fact that the name was arbitrarly used in places and I just replaced it in all the project even where it didn't have anything to do with the factory.

PR #521, opened 2024-07-12
@mintbase-js/rpc: add rpcUrl mandatory param

rubenmarcus:


package-lock.json
tifrel, line 1: I don't see any `package.json` changes, why did the lock file change?

packages/rpc/src/methods/balance.ts
```diff
@@ -12,10 +12,15 @@ export const getBalance = async (accountId: string, network?: Network, rpc?: RPC
...
+  }, network, rpc, rpcUrl);
+
   const balanceString = res?.result?.amount as string;
   if (!balanceString) {
     throw new Error(`Malformed response: ${JSON.stringify(res)}`);
   }
   return new BN(balanceString);
 };
+
+export const getBalance = async (accountId: string, network?: Network, rpc?: RPC_OPTIONS, rpcUrl?: string): Promise<BN> => {
```
tifrel: Why?

packages/rpc/src/methods/getBlockHash.ts
```diff
@@ -1,13 +1,13 @@
 import { Network } from '@mintbase-js/sdk';
 import { RPC_OPTIONS, requestFromNearRpc } from '../util';
 
-export const getBlockHash = async (network?: Network,  rpc?: RPC_OPTIONS): Promise<string> => {
+export const getBlockHash = async (network?: Network,  rpc?: RPC_OPTIONS, rpcUrl?: string): Promise<string> => {
```
tifrel:
```suggestion
export const getBlockHash = async (network?: Network, rpc?: RPC_OPTIONS, rpcUrl?: string): Promise<string> => {
```

packages/rpc/src/methods/account.ts
```diff
@@ -1,7 +1,7 @@
 import { Network } from '@mintbase-js/sdk';
 import { RPC_OPTIONS, requestFromNearRpc } from '../util';
 
-export const accountExists = async (accountId: string, network?: Network, rpc?: RPC_OPTIONS): Promise<boolean> => {
+export const accountExists = async (accountId: string, network?: Network, rpc?: RPC_OPTIONS, rpcUrl?: string): Promise<boolean> => {
```
tifrel: This opens up an edge case where someone could supply `network = 'mainnet'`, but a testnet RPC URL

packages/rpc/src/util.ts
```diff
@@ -2,19 +2,20 @@ import { mbjs, RPC_ENDPOINTS, NEAR_RPC_ENDPOINTS } from '@mintbase-js/sdk';
 import fetch from 'cross-fetch';
 
 
-export type RPC_OPTIONS  = 'lava' | 'near' | 'beta'
-
+export type RPC_OPTIONS  = 'lava' | 'near' | 'beta' | 'fastnear'
```
tifrel: Pagoda?

packages/rpc/src/methods/account.ts
bh2smith: Just a note that: 
The fact that we had to change the input parameters over so many files suggest that we should have a type interface for input arguments to our functions.

```
interface networkArgs {
   rpcUrl: string;
   network: Network;
}
```

So then we would only have to change it in one place.

bh2smith (commented):
lock file changes without dependency changes.

packages/rpc/src/util.ts
```diff
@@ -2,19 +2,20 @@ import { mbjs, RPC_ENDPOINTS, NEAR_RPC_ENDPOINTS } from '@mintbase-js/sdk';
...
 
 export const requestFromNearRpc = async (
   body: Record<string, unknown>,
   network?: string,
   rpc?:  RPC_OPTIONS,
-): Promise<{ result: Record<string, unknown>, error: unknown } | undefined> => {
+  rpcUrl?: string): Promise<{ result: Record<string, unknown>, error: unknown } | undefined> => {
 
   const fetchUrl =  mbjs.keys.nearRpcUrl || RPC_ENDPOINTS[mbjs.keys.rpc][mbjs.keys.network]  || NEAR_RPC_ENDPOINTS[mbjs.keys.network];
```
tifrel: I see 6 different branches to figure out a URL. We determine `fetchUrl` and `rpcAddress`, even if we might not need them. We give three different variable names to what is essentially the same thing. The suffix `Address` is usually used for smart contracts addresses.

This is a good opportunity to make this code less confusing.

packages/sdk/src/types.ts
```diff
@@ -68,6 +68,11 @@ export enum GRAPHQL_ENDPOINTS  {
   testnet = 'https://interop-testnet.hasura.app/v1/graphql',
 }
 
+export enum FASTNEAR_RPC_ENDPOINTS  {
+  mainnet = 'https://free.rpc.fastnear.com/',
+  testnet = 'https://g.w.lavanet.xyz:443/gateway/neart/rpc-http/f538cb3b0a85aafdb9996886d004ee0a',
+}
+
```
bh2smith, line 74: Isn't this a public project? We have hardcoded api keys here.

packages/sdk/src/types.ts
```diff
@@ -68,6 +68,11 @@ export enum GRAPHQL_ENDPOINTS  {
   testnet = 'https://interop-testnet.hasura.app/v1/graphql',
 }
 
+export enum FASTNEAR_RPC_ENDPOINTS  {
+  mainnet = 'https://free.rpc.fastnear.com/',
+  testnet = 'https://g.w.lavanet.xyz:443/gateway/neart/rpc-http/f538cb3b0a85aafdb9996886d004ee0a',
```
tifrel, line 72: That is a lava endpoint.

packages/sdk/src/types.ts
```diff
@@ -104,7 +109,7 @@ export type FtAddresses = {
   usdt: USDT_ADDRESS;
 }
 
-export type RPC_OPTIONS  = 'lava' | 'near' | 'beta'
+export type RPC_OPTIONS  = 'lava' | 'near' | 'beta' | 'fastnear'
```
tifrel, line 111: Why is this type duplicated here and in the rpc package?

tifrel (changes requested):

packages/sdk/src/types.ts
tifrel: Already existed before: https://github.com/Mintbase/mintbase-js/pull/521/files#diff-9d8c1fba9e81b9e94776fadecf0793c4bbacdf4474689eb6817576f5f7e6b5a8R76-R78

IIRC we did that knowing that we could deprecate the key and get a new one in case it gets abused.

packages/rpc/src/util.ts
tifrel: A way to simplify this is to have a single way to configure this: Using `mbjs.keys`.

packages/sdk/src/types.ts
sainthiago: fastnear doesn't have testnet.

packages/rpc/src/methods/ftMetadata.ts
```diff
@@ -16,6 +16,7 @@ interface FtMetadataProps {
   contractId: string;
   network?: Network;
```
tifrel, line 17:
```suggestion


```

packages/rpc/src/methods/ftMetadata.ts
```diff
@@ -16,6 +16,7 @@ interface FtMetadataProps {
...
   rpc?: RPC_OPTIONS;
```
tifrel:
```suggestion
```

packages/rpc/src/methods/ftStorageBalance.ts
```diff
@@ -6,15 +6,15 @@ interface FTStorageProps {
   accountId: string;
   network?: Network;
   rpc?: RPC_OPTIONS;
```
tifrel:
```suggestion
```

packages/rpc/src/methods/payouts.ts
```diff
@@ -35,9 +35,10 @@ interface  PayoutsProps {
   tokenId: string;
   network?: Network;
   rpc?: RPC_OPTIONS;
```
tifrel:
```suggestion
```

packages/rpc/src/methods/social.ts
```diff
@@ -46,12 +46,11 @@ const getImageUrl = (image: ProfileImage): string | null => {
 export const nearSocialProfile = async (
   accountId: string,
   network?: Network,
```
tifrel:
```suggestion
```

tifrel: I just now see we need this. Still a footgun.

tifrel:
@CodiumAI-Agent /review

packages/sdk/src/types.ts
rubenmarcus: because this was a fix, and I didnt wanted to have to reinstall sdk on rpc package just to have this new value on the type. we have the same problem with mintbase-wallet
https://github.com/Mintbase/mintbase-wallet/pull/1012

bh2smith:
Ok --- So this branch was never merged and our wallet is depending on it in production.

packages/rpc/src/util.ts
```diff
@@ -1,19 +1,56 @@
...
+type RpcNodes = {
+  [key in RPC_OPTIONS]?: string;
+};
+
+export const rpcNodes: {
+  mainnet: RpcNodes;
+  testnet: RpcNodes;
+} = {
+  mainnet: {
+    lava: "https://g.w.lavanet.xyz:443/gateway/near/rpc-http/f538cb3b0a85aafdb9996886d004ee0a",
```
bh2smith: Isn't this a public repo?

packages/rpc/src/util.ts
```diff
@@ -1,19 +1,56 @@
...
-    headers: { 'Content-type': 'application/json' },
-  });
 
-  return res.json();
+   const validUrls = Object.values(rpcNodes.mainnet).concat(Object.values(rpcNodes.testnet));
+
+  // Set default rpcUrl if not provided
+   if (!rpcUrl) {
+    rpcUrl = rpcNodes.mainnet.near;
+  }
```
bh2smith: Shouldn't we choose our default based on a networkId? This seems to fallback to mainnet if not provided.

package-lock.json
tifrel, line 1: Was this the same npm version?

rubenmarcus: react package needed a new install to build, i just did npm install on react package so it doesnt fail the build

packages/react/package.json
```diff
@@ -32,4 +32,4 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0"
   }
-}
+}
```
tifrel, line 36: Is this some tooling artifact? Saw the same change on a few files now.

packages/rpc/README.md
```diff
@@ -11,31 +11,31 @@ This module provides a wrapper around common RPC calls used to fetch on-chain da
 
 ## Available Modules & Methods
 
-`type Network: 'testnet' | 'mainnet'`
+We now accepts rpcUrl in each method, so you can pass your own RPC regarding one of those on the list:
```
tifrel:
```suggestion
We now accept `rpcUrl` on each method, so you can pass your own RPC regarding one of those on the list:
```

bh2smith (approved):
I am not caught up on the whole history of this change but it seems like a lot of similar changes are made in several places. While this doesn't prevent me from letting it pass for now, I think we can learn a lot from this and try to eliminate duplications by using type interfaces more effectively in the future. 

Note that 

1. It looks like we are leaking a lava rpc url, but I guess we don't mind.
2. The optional `rpcUrl` is a misleading convenience that falls back on mainnet if not supplied. We could add a comment to the method saying that this is what will happen. Unfortunately we would have to add this comment in ~30 different places until we have a common type interface that is used in all methods required similar parameters.

packages/rpc/README.md
```diff
@@ -11,31 +11,31 @@ This module provides a wrapper around common RPC calls used to fetch on-chain da
...
+`packages/rpc/src/util.ts#9`
 
-### `getBalance(account: string, network?: Network, rpc?: RPC_OPTIONS): BN`
+if **rpcUrl** not passed, it will fallback to near rpc mainnet.
```
tifrel:
```suggestion
If `rpcUrl` is not passed, it will fallback to the `https://rpc.mainnet.near.org`.
```

packages/rpc/src/util.ts
```diff
@@ -1,40 +1,69 @@
...
-    headers: { 'Content-type': 'application/json' },
-  });
+  // Set default rpcUrl if not provided
+   if (!rpcUrl) {
+    rpcUrl = rpcNodes.mainnet.near;
+  }
+
+  if (rpcUrl && !validUrls.includes(rpcUrl)) {
+    throw new Error('Invalid rpcUrl');
+  }
```
tifrel: Why should we curate a list here, what if someone wants to use their own private RPC?

packages/rpc/README.md
```diff
@@ -11,31 +11,31 @@ This module provides a wrapper around common RPC calls used to fetch on-chain da
 
 ## Available Modules & Methods
 
-`type Network: 'testnet' | 'mainnet'`
+We now accepts rpcUrl in each method, so you can pass your own RPC regarding one of those on the list:
+`packages/rpc/src/util.ts#9`
```
tifrel: Why would we keep this list?

packages/react/package.json
```diff
@@ -24,6 +24,7 @@
   },
   "dependencies": {
     "@mintbase-js/wallet": "0.6.1-update-packages-24058ee.0",
+    "@mintbase-js/data": "0.6.1-update-packages-24058ee.0",
```
bh2smith, line 27: Just a thought, we could use this PR to find and replace all of `0.6.1-update-packages-24058ee.0` with with `^0.6.1`

packages/rpc/src/methods/ftStorageBalance.ts
```diff
@@ -1,20 +1,17 @@
...
 
-export const ftStorageBalance = async ({ contractId, accountId, network, rpc }: FTStorageProps): Promise<string | null> => {
+export const ftStorageBalance = async ({ contractId, accountId,  rpcUrl }: FTStorageProps): Promise<string | null> => {
   const balance = await callViewMethod<{ total: string; available: string } | null>({
     contractId,
     method: 'storage_balance_of',
     args: { account_id: accountId },
-    network: network,
-    rpc: rpc,
+    rpcUrl: rpcUrl,
```
tifrel:
```suggestion
    rpcUrl,
```

packages/rpc/src/methods/ftMetadata.ts
```diff
@@ -52,12 +50,11 @@ function isStringOrNull(x: any): x is string | null {
...
 }
 
-export const ftMetadata = async ({ contractId, network, rpc }: FtMetadataProps): Promise<FtMetadata | null> => {
+export const ftMetadata = async ({ contractId, rpcUrl }: FtMetadataProps): Promise<FtMetadata | null> => {
   const res = callViewMethod<FtMetadata>({
     contractId,
     method: 'ft_metadata',
-    network: network,
-    rpc: rpc,
+    rpcUrl: rpcUrl,
```
tifrel:
```suggestion
    rpcUrl,
```

packages/rpc/src/util.ts
rubenmarcus: replaced lava rpc url with the one on their docs

packages/rpc/src/util.ts
```diff
@@ -51,13 +80,13 @@ export const callViewMethod = async <T>({
...
 
   if (res?.error) {
     throw res.error;
   }
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore
-  const parsed = JSON.parse(Buffer.from(res?.result?.result).toString());
+
+  const resultBuffer = Buffer.from(res?.result?.result as string);
+  const parsed = JSON.parse(resultBuffer.toString());
```
tifrel, line 69: Is this a conversion string -> buffer -> string?

packages/rpc/src/util.ts
rubenmarcus: yes but then I need to accept networkId somewhere, maybe in every method then?

packages/sdk/src/types.ts
```diff
@@ -67,6 +67,11 @@ export enum GRAPHQL_ENDPOINTS  {
   testnet = 'https://interop-testnet.hasura.app/v1/graphql',
 }
 
+export enum FASTNEAR_RPC_ENDPOINTS  {
```
tifrel, line 70: Again, we wouldn't need to curate lists here. This also resolves the leakage of our private lava endpoints.

packages/storage/src/node-uploads.ts
```diff
@@ -7,7 +7,6 @@ import {
   MAX_UPLOAD_BYTES,
   MAX_UPLOAD_ERROR_MSG,
   MINTBASE_API_ANON_USER,
-  MINTBASE_API_KEY_HEADER,
```
tifrel, line 10: Fixing the test but removing it here? What's the background on this?

tifrel (changes requested):

packages/rpc/src/util.ts
```diff
@@ -1,40 +1,50 @@
...
 import fetch from 'cross-fetch';
 
 
-export type RPC_OPTIONS  = 'lava' | 'near' | 'beta'
+
+export const queryNearRpc = async ({params, method, rpcUrl}):  Promise<{ result: Record<string, unknown>, error: unknown } | undefined> => {
+  return await requestFromNearRpc({
+    jsonrpc: '2.0',
+    id: 'dontcare',
+    method:method,
```
tifrel, line 7: If the name here is `queryNearRpc`, why don't we fill in `method: 'query'` and save typing on usage?

rubenmarcus: because its just a wrapper and we use different methods on the calls

PR #522, opened 2024-07-14
fix fetch token price to Binance API

rubenmarcus:
promise.any calls both apis at same type and coingecko has ratelimits that make it call many times with error.. Binance always give 200.

Screenshot 2024-07-14 at 23 18 06

### DIFF
--- a/packages/data/src/api/nearPrice/nearPrice.ts
+++ b/packages/data/src/api/nearPrice/nearPrice.ts
@@ -22,7 +22,6 @@ export const nearPrice = async (): Promise<ParsedDataReturn<string>> => {
         return await fetchPrice<CoinGeckoNearPriceData>(COIN_GECKO_API, data => ({ price: data.near.usd }));
       });
 
-    console.log(res, 'res');
     return parseData(res.price);
   } catch (err) {
     console.error(
--- a/packages/data/src/api/nearPrice/nearPrice.ts
+++ b/packages/data/src/api/nearPrice/nearPrice.ts
@@ -16,12 +16,13 @@ async function fetchPrice<T>(url: string, tokenPrice: (data: T) => NearPriceData
 
 export const nearPrice = async (): Promise<ParsedDataReturn<string>> => {
   try {
-    let res: NearPriceData;
-    try {
-      res = await fetchPrice<NearPriceData>(BINANCE_API, data => data);
-    } catch (err) {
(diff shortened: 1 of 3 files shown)


bh2smith (changes requested):
While the proposed solution does eliminate console errors, it does so by swallowing errors and requests from the different price feeds asymmetrically. So when we do see the rate limit errors (which we will) we will only ever see the BinanceAPI error. 

I think this problem could be addressed more appropriately with some basic caching. Our servicies shouldn't be requesting the price of near more than once every minute or two. I would propose using something like https://www.npmjs.com/package/cache-manager with an expiry of 120 seconds.

Oh - I just realized that this is mintbase-js. The caching should be done on the application layer.

sainthiago (approved):

rubenmarcus:
@bh2smith 
`
The caching should be done on the application layer.`

you already answered yourself.

**Binance** doesnt have rate limit, as I said on desc. **Coingecko** has.

packages/data/src/api/nearPrice/nearPrice.ts
```diff
@@ -8,25 +8,20 @@ import {
...
-      fetchPriceFromCoinGecko(),
-      fetchPriceFromBinance(),
-    ]);
-
+    let res: NearPriceData;
+    try {
+      res = await fetchPrice<CoinGeckoNearPriceData>(COIN_GECKO_API, data => ({ price: data.near.usd }));
+    } catch (err) {
+      res =  await fetchPrice<NearPriceData>(BINANCE_API, data => data);
+    }
```
bh2smith: You are prioritizing CoinGecko by calling it first and swallowing their errors. 

These two APIS are different. Just keep the Promise.any, catch rate limit errors and handle them differently (like return null). 

Have you considered adding another API? 

https://coinmarketcap.com/api/pricing/

https://coinpaprika.com/api/

rubenmarcus:
```
Have you considered adding another API?

https://coinmarketcap.com/api/pricing/

https://coinpaprika.com/api/
```

this is something I needed to do later, user can pass it api and we fetch for him.

rubenmarcus: Screenshot 2024-07-15 at 11 00 11

bh2smith (approved):

